PR #25, opened 2021-04-28
AboutOrganisationController unit tests

kiran-mastek:
Complete class converage
Add mappings for models with tests

### DIFF
--- a/src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Models/BuyingCatalogue/CatalogueItem.cs
+++ b/src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Models/BuyingCatalogue/CatalogueItem.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 
 #nullable disable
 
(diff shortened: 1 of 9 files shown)


src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Models/BuyingCatalogue/CatalogueItem.cs
```diff
@@ -26,5 +27,10 @@ public CatalogueItem()
         public virtual AssociatedService AssociatedService { get; set; }
         public virtual Solution Solution { get; set; }
         public virtual ICollection<CataloguePrice> CataloguePrices { get; set; }
+
+        public MarketingContact FirstContact() => Solution?.MarketingContacts?.FirstOrDefault() ?? new MarketingContact();
+
+        public MarketingContact SecondContact() =>
+            Solution?.MarketingContacts?.Skip(1).FirstOrDefault() ?? new MarketingContact();
     }
```
MichaelRobinson-BJSS, line 35: These EF classes were reverse generated from the current databases. If the databases change between now and migration completion I may need to regenerate them. As the classes are generated as partial, could you put any manual changes into secondary file?

kiran-mastek: Good point - done.

PR #150, opened 2021-06-23
Consolidate databases

bjss-chris:
Activity: AB#14974
Task: AB#15186

Please note: further work is required.

Catalogue solution and additional services journeys will be broken until the typing is corrected – intend to pick this up next.


src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Extensions/CatalogueItemTypeExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+                case CatalogueItemType.AdditionalService:
+                    order.Progress.AdditionalServicesViewed = true;
+                    break;
+
+                case CatalogueItemType.AssociatedService:
+                    order.Progress.AssociatedServicesViewed = true;
+                    break;
+
+                default:
+                    throw new ArgumentException($"{nameof(itemType)} must be a valid {nameof(CatalogueItemType)}", nameof(itemType));
```
PhysicsJon, line 43: What benefits do we have for the use of a switch statement here rather than a switch expression as used elsewhere?

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Models/GPITBuyingCatalogue/CatalogueItemType.cs
```diff
@@ -1,15 +1,16 @@
...
+
+        [Display(Name = "Additional Service")]
+        AdditionalService = 2,
 
-        public CatalogueItemType(int id, string name)
-            : base(id, name)
-        {
-        }
+        [Display(Name = "Associated Service")]
+        AssociatedService = 3,
```
PhysicsJon, line 14: I believe when you set the first item in an enum to a number index, the rest follow on without needing explicit indexing. Would this make some of the enums a little simpler?

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Extensions/CatalogueItemTypeExtensions.cs
bjss-chris: I can't use a switch expression here because I'm just executing statements such as `order.Progress.CatalogueSolutionsViewed = true;` rather than calculating a value or returning as value, like in `InferEstimationPeriod`.

Therefore I'd probably have to change the implementation to return a lambda but that I'd then need to make other changes. Personally, I think the switch statement is fine here but we can always review again...

Give me a thumbs up if you're satisfied with my response and I'll resolve the conversation.

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Models/GPITBuyingCatalogue/CatalogueItemType.cs
bjss-chris: You are correct that the rest will follow on. I just copied these over from the ordering API as they're defined there. We could remove the explicit indexing after the first value although I'm not sure if it's worth doing right now, as we should apply the same style to all enums.

Maybe we add to tech debt to review preferred style for defining enums?

PhysicsJon: A tech debt ticket to review would be a decent way forwards, I think we can set this conversation to resolved

PhysicsJon (approved):

PR #155, opened 2021-06-25
Fix typing of call-off ID and catalogue item ID

bjss-chris:
Activity: AB#14974
Task: AB#15186

Similar to the previous PR, there are still tasks to do following this PR.

Some unit tests will need to be reviewed. There are also some places where the nullability of a value type probably needs to be reviewed. There is also so much code that needs tidying up (lazy lambdas every where, badly written unit tests etc.) that I have done bits and pieces as part of this PR but not consistently – there is just far too much and this is already a large PR. I will create follow-up PRs for remaining issues.


PhysicsJon (changes requested):
a few small changes to add consistency

PhysicsJon (approved):

PR #159, opened 2021-06-25
Move all unit tests to xUnit

bjss-chris:
Activity: AB#14974
Task: AB#15485

Some disabled tests to fix up after (but not a huge number).


tests/NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests/Areas/Marketing/MappingProfiles/BrowserBasedProfileTests.cs
```diff
@@ -50,13 +47,12 @@ public void SetUp()
...
 
-        [OneTimeTearDown]
-        public void CleanUp()
+        public void Dispose()
         {
             mapper = null;
         }
 
-        [TearDown]
+        // TODO: No tear down in xUnit
```
PhysicsJon, line 55: You can do the teardown in dispose methods

PhysicsJon (approved):
A couple of suggestions only.

You have a fair number of todo's regarding setup/teardown parts not included in xUnit. The best way I've found of setting these up is to implement IClassFixture<T> with a setup class that can be injected into the constructor of the test class and make sure that the invoked class implements IDisposable to do the teardown. It's not as clean, but it separates the setup to a specific class

bjss-chris:
> A couple of suggestions only.
> 
> You have a fair number of todo's regarding setup/teardown parts not included in xUnit. The best way I've found of setting these up is to implement IClassFixture with a setup class that can be injected into the constructor of the test class and make sure that the invoked class implements IDisposable to do the teardown. It's not as clean, but it separates the setup to a specific class

Good points. Let's get it in and then follow up on your points.

PR #164, opened 2021-06-28
Speed up unit tests

kiran-mastek:
Replace customisations slowing unit tests start with custom Fakers
Correct check for `AdditionalService` in `CatalogueItem`
Re-add tests ignored earlier by using `Xunit` constructor to replace `NUnit` `SetUp` method.


bjss-chris:
I'd rather not have two different approaches for generating unit test data. My preference would be to stick with AutoFixture.

It means we'll have to suffer the slower tests for another day or two, but I'd like to postpone merging this PR until we've had more time to look into improving the speed with AutoFixture.

If you want to, feel free to create a separate PR for your unit test fixes (not essential though).

PR #199, opened 2021-07-07
Remove search bar

RahulJha-mastek:
Story AB#15247
Task AB#15460


leerogers-bjss (approved):
hey rahul, just for next time, don't add css classes directly to main. you'll need to add them in as a seperate scss file under wwwroot > scss. if you don't then your changes would be lost because main.css gets rebuilt in the build pipeline.

beyond that, approved

PR #221, opened 2021-07-08
Delete Webapp.Integrations.Tests Project

leerogers-bjss:
Talked to jon, this is no longer needed due to the existance of the E2E Tests


bjss-chris (changes requested):
We might want to add some integration tests, rather than E2E tests, at some point. However, there's no point in having an empty project lying around either so fine with the delete.

We'll need to update the pipeline as well though as that has a step for running integration tests.

MichaelRobinson-BJSS:
Updated the build pipeline to remove integration tests in this PR https://github.com/nhs-digital-gp-it-futures/GPITBuyingCatalogue/pull/236

bjss-chris (approved):

PR #235, opened 2021-07-09
Removed concept of private instance

MichaelRobinson-BJSS:
Removed concept of private instance and added admin auth to marketing editing pages


PhysicsJon (commented):
E2E sections look good, can't see anything else wrong with the rest of the changes, but would like @bjss-chris or @leerogers-bjss to take a glance over when they have time

leerogers-bjss (approved):

PR #249, opened 2021-07-14
Fix Layout and Add a Solution Issues

leerogers-bjss:
Move all Layout files to a central location and combine them into understandable layouts and partials.
Update all layout files to use the same, new footer
update all layout files to use new menu look, with the separation being either admin or non-admin menu.
Remove duplicate HTML into different partials which are then called by the layout pages
update all _ViewStart files to reference these layout files
update pages to call the correct layout files.

Fix add a solution to attach the error correctly to the list of checkboxes by fixing the fluid validation property attachment

update add a solution to return the catalogueItemId so that we can then redirect to the manage solutions page after creating a new solution


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Views/AddCatalogueSolution/Index.cshtml
```diff
@@ -50,9 +49,14 @@
                                                   hidden-input="@Model.Frameworks[i].FrameworkId"
                                                   data-test-id="framework-names" />
                                 }
+
                             }
                         </nhs-checkbox-container>
                     </nhs-fieldset-form-label>
+                    @for (var i = 0; i < Model.Frameworks.Count; i++)
```
PhysicsJon, line 56: could this be done as a foreach?

PhysicsJon (changes requested):
one small change

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Views/AddCatalogueSolution/Index.cshtml
leerogers-bjss: nope, needs the correct list reference or it won't attach correct on the post

PhysicsJon (approved):

PR #296, opened 2021-08-03
Feature/15984 interoperability view

MichaelRobinson-BJSS:


src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsService.cs
```diff
@@ -268,6 +268,15 @@ public async Task SaveSolutionDescription(CatalogueItemId solutionId, string sum
             await solutionRepository.SaveChangesAsync();
         }
 
+        public async Task SaveInteroperabilityLink(CatalogueItemId solutionId, string link)
+        {
+            link.ValidateNotNullOrWhiteSpace(nameof(link));
+
+            var solution = await solutionRepository.SingleAsync(s => s.Id == solutionId);
```
PhysicsJon, line 275: Do we need to validate that the solution is returned correctly?

PhysicsJon (changes requested):

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsService.cs
MichaelRobinson-BJSS: It's a SingleAsync so if it doesn't exist it will throw an exception anyway. I think an error page is appropriate in this situation.

PhysicsJon (approved):

PR #314, opened 2021-08-10
Correctly set back link

MichaelRobinson-BJSS:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/BrowserBasedModels/BrowserBasedModel.cs
```diff
@@ -15,6 +15,7 @@ public BrowserBasedModel()
         public BrowserBasedModel(CatalogueItem catalogueItem)
             : base(catalogueItem)
         {
+            BackLink = $"/admin/catalogue-solutions/manage/{catalogueItem.Id}/client-application-type/add-application-type";
```
PhysicsJon, line 18: According to the acceptance criteria, this is correct, however there is an instance where this is not correct. If you Edit an already added Browser Based client application type this goes back to the wrong page. @bjss-chris is there anything we could do to satisfy both cases?

MichaelRobinson-BJSS: Updated to deal with Edit scenario. However if you add a new browser-based then edit something it creates a browser-based client application type so the back button doesn't go back to 'add client application'. I think that should be acceptable given its internal use. Alternative is to have state or query strings which sounds over the top to me.

PhysicsJon (approved):

PR #331, opened 2021-08-13
Convert user ID to int

bjss-chris:
Activity: AB#14974
Task: AB#15406


tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Utils/RandomData/GenerateUser.cs
```diff
@@ -20,7 +20,7 @@ internal static User Generate()
         internal static AspNetUser GenerateAspNetUser(int organisationId, string password, bool isEnabled)
         {
             var user = new Faker<AspNetUser>("en_GB")
-                .RuleFor(u => u.Id, f => f.Random.Guid())
+                .RuleFor(u => u.Id, f => f.Random.Int())
```
PhysicsJon, line 23: Not sure this rule is even required any more (due to the table column being an Identity column)

PhysicsJon (dismissed):
Approved, but I think there is a little change that would be nice to do

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Utils/RandomData/GenerateUser.cs
bjss-chris: I'll give it a test.

PhysicsJon (approved):

PR #342, opened 2021-08-18
Feature/15981 add desktop app type

MichaelRobinson-BJSS:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Controllers/CatalogueSolutionsController.cs
```diff
@@ -540,6 +540,10 @@ public async Task<IActionResult> AddApplicationType(CatalogueItemId solutionId,
                     model.SelectedApplicationType.ToString(),
                     typeof(MobileTabletBasedController).ControllerName(),
                     new { solutionId }),
+                ServiceContracts.Solutions.ClientApplicationType.Desktop => RedirectToAction(
+                    model.SelectedApplicationType.ToString(),
```
leerogers-bjss, line 544: I'm not sure i like this ToString
The reason we moved to nameof(_controllermethod_) is so that we wouldn't be using strings as part of a redirect to action (and to keep a single point of truth for the route values)

i see what you're trying to do, but we have examples of doing if-else and switches to select different redirect to actions pointing to different controllermethods.

so i would change it to that :)

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/InteroperabilityModels/AddGpConnectIntegrationModel.cs
```diff
@@ -33,15 +33,15 @@ public AddGpConnectIntegrationModel(CatalogueItemId solutionId)
 
         public IEnumerable<object> IntegrationTypes { get; }
 
-        [Required(ErrorMessage = "Select an integration type")]
+        [Required(ErrorMessage = "Select integration type")]
         public string SelectedIntegrationType { get; set; }
 
-        [Required(ErrorMessage = "Select Provider or Consumer")]
+        [Required(ErrorMessage = "Select if your system is a provider or consumer")]
```
leerogers-bjss, line 39: has this been signed off by luke?

only thing is NHS prefer simple and to the point error messages (which always seem super snarky to me) and the "Select provider or consumer" seems more inline with what they would want :)

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/InteroperabilityModels/AddIm1IntegrationModel.cs
```diff
@@ -33,15 +33,15 @@ public AddIm1IntegrationModel(CatalogueItemId solutionId)
 
         public IEnumerable<object> IntegrationTypes { get; }
 
-        [Required(ErrorMessage = "Select an integration type")]
+        [Required(ErrorMessage = "Select integration type")]
         public string SelectedIntegrationType { get; set; }
 
-        [Required(ErrorMessage = "Select Provider or Consumer")]
+        [Required(ErrorMessage = "Select if your system is a provider or consumer")]
```
leerogers-bjss, line 39: same as above

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Views/DesktopBased/Desktop.cshtml
```diff
@@ -0,0 +1,100 @@
...
+    </div>
+</div>
+
+
+@{
+    NhsTagsTagHelper.TagColour GetTagColour(string status) =>
+        status switch
+        {
+            "COMPLETE" => NhsTagsTagHelper.TagColour.Green,
+            "INCOMPLETE" => NhsTagsTagHelper.TagColour.Grey,
```
leerogers-bjss, line 97: We're not using Incomplete anymore

the options available are:
- Complete
- Not Started
- Cannot Start Yet
- Optional
- In Progress

so this should be Not Started
Also it should be only first letter capitalised

leerogers-bjss (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/InteroperabilityModels/AddGpConnectIntegrationModel.cs
MichaelRobinson-BJSS: This is what is in the acceptance criteria and was even raised as a bug. I don't recall seeing it in the AC when I first looked at this so I initially used my own judgement which I agree appears better, but the AC is the AC. It can easily be changed later though.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/InteroperabilityModels/AddIm1IntegrationModel.cs
MichaelRobinson-BJSS: This is what is in the acceptance criteria and was even raised as a bug. I don't recall seeing it in the AC when I first looked at this so I initially used my own judgement which I agree appears better, but the AC is the AC. It can easily be changed later though.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Views/DesktopBased/Desktop.cshtml
MichaelRobinson-BJSS: Done. As per previous discussions, the other states will be dealt with in an additional story.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Controllers/CatalogueSolutionsController.cs
MichaelRobinson-BJSS: Agreed. Was just a continuation of what Neha had done.

leerogers-bjss (approved):

PR #389, opened 2021-09-24
Remove automapper from solution features

PhysicsJon:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/SolutionFeaturesModel.cs
```diff
@@ -2,7 +2,12 @@
 {
     public class SolutionFeaturesModel : SolutionDisplayBaseModel
     {
-        public string[] Features { get; set; }
+        public SolutionFeaturesModel(string[] features)
```
bjss-colin, line 5: Do we need a parameterless ctor?

PhysicsJon: It's not used anywhere else, and only has read only properties so I don't believe we do

bjss-colin (approved):

PR #396, opened 2021-09-27
Remove AutoMapper (hosting types)

bjss-chris:
The extension method approach (for the `Status` method) may not be the best approach long-term but it was the simplest option to keep this PR from becoming too large.


src/NHSD.GPIT.BuyingCatalogue.Framework/Extensions/HybridHostingTypeExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using NHSD.GPIT.BuyingCatalogue.EntityFramework.Catalogue.Models;
+using NHSD.GPIT.BuyingCatalogue.ServiceContracts.Enums;
+using NHSD.GPIT.BuyingCatalogue.ServiceContracts.Solutions;
+
+namespace NHSD.GPIT.BuyingCatalogue.Framework.Extensions
+{
+    public static class HybridHostingTypeExtensions
+    {
+        public static TaskProgress Status(this HybridHostingType hybridHostingType)
```
bjss-colin, line 9: What about a base class, for the available hosting types, that implements this `Status` method?

It would need to be overridden once for `PublicCloud`

bjss-chris: I think this thought had crossed my mind previously but didn't on this occasion for some reason, but sounds like a sensible suggestion!

Will look at doing that.

bjss-chris: Actually, I can't do this (easily, anyway) because those extension methods all return `TaskProgress`, which is defined in the `ServiceContracts` project.

We could move `TaskProgress` but it's used in a lot of places, which is will increase the size of the PR quite a bit (even if it's just namespace changes) so I suggest we look at doing something like this separately.

bjss-colin: Yeah that's fine, we can look at this in the future as a piece of tech debt when we have defined the scope a bit better so as to not over-extend the purpose of this PR.

bjss-colin (approved):

PR #405, opened 2021-09-29
Fix base model of all view models

PhysicsJon:
Fixes an issue where some pages were not shown in the list of pages
Fixes an issue with some pages not showing pagination links
Adds unit tests for ToPricingUnitString


src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Models/CataloguePrice.cs
```diff
@@ -44,7 +44,7 @@ public override string ToString()
 
         public string ToPriceUnitString()
         {
-            return $"{PricingUnit.Description} {TimeUnit.Value.Description()}".Trim();
+            return $"{PricingUnit?.Description ?? string.Empty} {(TimeUnit.HasValue ? TimeUnit.Value.Description() : string.Empty)}".Trim();
```
bjss-colin, line 47: `string.Join` is also an option here as it would automatically trim if `Description`` or `TimeUnit` are null/empty.

This comes down to personal preference however so it's at your discretion

bjss-colin (approved):

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Models/CataloguePrice.cs
PhysicsJon: I prefer the interpolation method as it's very explicit in what it does. We can have a conversation about this when we have a code standards meeting at some point in the future

PR #406, opened 2021-09-29
Add typescript

MichaelRobinson-BJSS:


bjss-colin:
Is this ready to be published? It may fall behind with the number of PRs being merged to main.

bjss-chris (approved):
It looks like we're not actually using any TS functionality in our scripts at the moment but I guess the purpose of this PR is just to facilitate that.

leerogers-bjss (approved):
so just had a look

since the TS change doesn't affect JS functionality, i'm happy for this to go ahead.

the conversion from JS to TS on Filtering and ManageSolutions JS can be done once this is in.

PR #411, opened 2021-09-30
Feature/remove newtonsoft

MichaelRobinson-BJSS:
Removed remaining references to Newtonsoft


src/NHSD.GPIT.BuyingCatalogue.Framework/Extensions/SolutionExtensions.cs
```diff
@@ -12,19 +12,19 @@ public static ClientApplication GetClientApplication(this Solution solution)
             if (string.IsNullOrWhiteSpace(solution.ClientApplication))
                 return new ClientApplication();
 
-            return JsonSerializer.Deserialize<ClientApplication>(solution.ClientApplication);
+            return JsonSerializer.Deserialize<ClientApplication>(solution.ClientApplication, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
```
PhysicsJon, line 15: I think the number of times JsonSerializerOptions is added to calls would suggest it should be split out to call in everywhere in a single definition

PhysicsJon (changes requested):
Looks good, but I've added a comment that would have wide reaching implications

src/NHSD.GPIT.BuyingCatalogue.Framework/Extensions/SolutionExtensions.cs
bjss-colin: +1. A static method somewhere would be beneficial. It would make any future changes far easier as we'd have to update only a single source.

MichaelRobinson-BJSS: Fair point, but does it need doing as part of this PR?

bjss-chris: I'd be tempted to go with something like this.

```c#
public static class JsonDeserializer
{
    public static T Deserialize<T>(string json)
    {
        return JsonSerializer.Deserialize<T>(json, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
    }
}
```

Feel free to think of better names for the class and method...

MichaelRobinson-BJSS: Ok, sigh...

PR #415, opened 2021-09-30
Add Suspended Solution Notice

leerogers-bjss:
Add Suspended Solution Notice to Description Page and Hide the Menu/Pagination

Add Logic to all the Public Browse Solution Endpoints to redirect back to description if you attempt to go to them directly and the solution is suspended.

Update some of the Public Browse Models to no longer take in a Solution but to instead take in the CatalogueItem itself (was silly to have two paths)

Add E2E Tests

### DIFF
--- a/tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/PublicBrowse/Solution/SupplierDetails.cs
+++ b/tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/PublicBrowse/Solution/SupplierDetails.cs
@@ -1,16 +1,19 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.EntityFrameworkCore;
 using NHSD.GPIT.BuyingCatalogue.E2ETests.Utils;
 using NHSD.GPIT.BuyingCatalogue.E2ETests.Utils.TestBases;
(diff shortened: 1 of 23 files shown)


tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/PublicBrowse/Solution/SupplierDetails.cs
```diff
@@ -46,5 +49,30 @@ public async Task SupplierDetails_VerifySupplierSummary()
...
+                    nameof(SolutionsController.Description))
+                .Should()
+                .BeTrue();
+        }
+
+        public void Dispose()
+        {
+            using var context = GetEndToEndDbContext();
+            context.CatalogueItems.Single(ci => ci.Id == SolutionId).PublishedStatus = PublicationStatus.Published;
+            context.SaveChanges();
```
PhysicsJon, line 75: The resetting of the published status could be refactored to a method on the test base that can be called here rather than repeating this every time

PhysicsJon (approved):
I have suggestions that would not necessitate a request for changes, but would be good to look into in future refactors

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/PublicBrowse/Solution/SupplierDetails.cs
leerogers-bjss: yeah, prob worth a look

PR #416, opened 2021-09-30
Add publication status automated tests

bjss-colin:


tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/Admin/ManageCatalogueSolution.cs
```diff
@@ -0,0 +1,172 @@
...
+            CommonActions
+                .GetRadioButtonsOptions()
+                .Should()
+                .BeEquivalentTo(expectedPublicationStatuses.Select(p => p.Description()));
+        }
+
+        [Fact]
+        public async Task ManageCatalogueSolution_SetPublicationStatus()
+        {
+            await using (var context = GetEndToEndDbContext())
```
PhysicsJon, line 136: You can simplify the using statement here (see above test)

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/Admin/ManageCatalogueSolution.cs
```diff
@@ -0,0 +1,172 @@
...
+            CommonActions
+                .PageLoadedCorrectGetIndex(
+                    typeof(CatalogueSolutionsController),
+                    nameof(CatalogueSolutionsController.Index))
+                    .Should()
+                .BeTrue();
+
+            await using (var context = GetEndToEndDbContext())
+            {
+                var publishedStatus = (await context.CatalogueItems.SingleAsync(c => c.Id == SolutionId)).PublishedStatus;
```
PhysicsJon, line 158: as above

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/Admin/ManageCatalogueSolution.cs
```diff
@@ -0,0 +1,172 @@
...
+            CommonActions
+                .ClickRadioButtonWithText(PublicationStatus.Published.Description());
+
+            CommonActions.ClickSave();
+
+            CommonActions
+                .PageLoadedCorrectGetIndex(
+                    typeof(CatalogueSolutionsController),
+                    nameof(CatalogueSolutionsController.Index))
+                    .Should()
```
PhysicsJon, line 153: Indentation has gone a bit weird

PhysicsJon (changes requested):

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/Admin/ManageCatalogueSolution.cs
bjss-colin: The reason for this is because it's used twice in the same test to confirm the publication status is being set when the user clicks save.

PhysicsJon: if you just do `await using var context = GetEndToEndDbContext();` it stays open until all calls are completed then disposes of the context. You don't need to initialize it again

bjss-colin: Introduced a new variable. Entity Framework was caching `CatalogueItem` on the first query which is why the context was being disposed as soon as it finished, and also to re-use the same variable name.

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/Admin/ManageCatalogueSolution.cs
```diff
@@ -150,16 +148,14 @@ await using (var context = GetEndToEndDbContext())
...
                 .BeTrue();
 
-            await using (var context = GetEndToEndDbContext())
-            {
-                var publishedStatus = (await context.CatalogueItems.SingleAsync(c => c.Id == SolutionId)).PublishedStatus;
-                publishedStatus
-                    .Should()
-                    .Be(PublicationStatus.Published);
-            }
+            await using var updatedContext = GetEndToEndDbContext();
```
PhysicsJon, line 154: you can remove this line and set the below calls to context as it still reads from the db regardless of any external updates via the application

PhysicsJon (changes requested):

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/Admin/ManageCatalogueSolution.cs
bjss-colin: That's not the case here. When using `context` in the second instance it returns the Draft Publication Status from the beginning of the test rather than fetching the value that was updated in the application.

image

The context needs to be disposed of and recreated for it to retrieve the actual value from the data store and not the cache.

PhysicsJon: Fair, I'm going to investigate further about this as it doesn't seem right.

PhysicsJon (approved):

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/Admin/ManageCatalogueSolution.cs
bjss-colin: There's some [more info here](https://stackoverflow.com/a/68268995/5062791) with one way on how to disable that behaviour

PR #424, opened 2021-10-06
Addressed TODO comments

MichaelRobinson-BJSS:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Startup.cs
```diff
@@ -129,12 +129,10 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
...
                         {
                             var error = context.Features.Get<IExceptionHandlerFeature>();
+
                             if (error != null)
                             {
-                                var errorMessage = error.Error.FullErrorMessage();
-
-                                // TODO - AppInsights isn't picking up LogError for some reason
-                                logger.LogInformation(error.Error, errorMessage);
+                                logger.LogError(error.Error, error.Error.FullErrorMessage());
```
bjss-chris, line 135: I'm not sure about this... I think that rather than passing `FullErrorMessage` we should use a message template, e.g. `"Exception details: {exception}"` (not intended as a good example of a message) and pass any arguments separately as `args`, so we can make full use of structured logging.

bjss-chris (commented):
Looks fine overall – just one thing I think we ought to consider changing around logging.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Startup.cs
```diff
@@ -128,13 +128,11 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
...
-                            if (error != null)
-                            {
-                                var errorMessage = error.Error.FullErrorMessage();
+                            var exceptionHandlerFeature = context.Features.Get<IExceptionHandlerFeature>();
 
-                                // TODO - AppInsights isn't picking up LogError for some reason
-                                logger.LogInformation(error.Error, errorMessage);
+                            if (exceptionHandlerFeature != null)
+                            {
+                                logger.LogError($"Exception occured: {exceptionHandlerFeature.Error}");
```
bjss-chris, line 135: Sorry, rather than string interpolation I meant like this:

```c#
logger.LogError("Exception occured: {Error}", exceptionHandlerFeature.Error);
```

This way the logging framework has a reference to the error object.

bjss-chris (changes requested):

bjss-chris (approved):

PR #428, opened 2021-10-06
Removed orderitemservice validation

MichaelRobinson-BJSS:


src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Models/OrderItemRecipientModel.cs
```diff
@@ -31,12 +32,12 @@ public OrderItemRecipientModel(ServiceRecipient recipient)
             {
                 if (!deliveryDate.HasValue)
                 {
-                    if (int.TryParse(Year, out int yearInt)
-                        && int.TryParse(Month, out int monthInt)
-                        && int.TryParse(Day, out int dayInt))
+                    try
+                    {
+                        deliveryDate = DateTime.ParseExact($"{Day}/{Month}/{Year}", "d/M/yyyy", CultureInfo.InvariantCulture);
```
bjss-colin, line 37: Could `Day`, `Month` and `Year` potentially be `int?`/`Nullable<int>`? This would allow us to use `new DateTime(Year, Month, Day)`

Alternatively, `TryParseExact` might fit better here.

bjss-colin (approved):
I've marked as approved with one comment for consideration. I appreciate this is a tech debt PR focusing on one aspect so it may be more desirable to create a separate tech debt item if you think my point is valid.

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Models/OrderItemRecipientModel.cs
MichaelRobinson-BJSS: I will review under another PR. Need to think about it. It becomes trickier with ints because they want it 0 padded in the view so one way or another they need converting to strings.

PR #433, opened 2021-10-08
Add navigation properties for CatalogueItemContacts

bjss-colin:
Adds the two-way navigation properties on `CatalogueItem` and `SupplierContact` for a many to many relationship

### DIFF
--- a/src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Configuration/CatalogueItemEntityTypeConfiguration.cs
+++ b/src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Configuration/CatalogueItemEntityTypeConfiguration.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using NHSD.GPIT.BuyingCatalogue.EntityFramework.Catalogue.Models;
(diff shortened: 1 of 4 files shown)


src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsService.cs
```diff
@@ -48,8 +48,9 @@ public Task<CatalogueItem> GetSolution(CatalogueItemId solutionId)
         {
             return dbContext.CatalogueItems
                 .Include(i => i.Solution)
+                .Include(i => i.CatalogueItemContacts)
                 .Include(i => i.CatalogueItemCapabilities).ThenInclude(sc => sc.Capability)
-                .Include(i => i.Supplier)
+                .Include(i => i.Supplier).ThenInclude(i => i.SupplierContacts)
```
PhysicsJon, line 53: I think the lambda param for the `ThenInclude` could be better named

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Configuration/CatalogueItemEntityTypeConfiguration.cs
```diff
@@ -40,6 +41,13 @@ public void Configure(EntityTypeBuilder<CatalogueItem> builder)
                 .OnDelete(DeleteBehavior.ClientSetNull)
                 .HasConstraintName("FK_CatalogueItems_Supplier");
 
+            builder.HasMany(i => i.CatalogueItemContacts)
+                .WithMany(s => s.AssignedCatalogueItems)
+                .UsingEntity<Dictionary<string, object>>(
```
bjss-chris, line 46: How does the `Dictionary<string, object>` work here? Will a type more specific than `object` work here? Also, what does EF use for the key?

I think this is a fairly new feature and I've not used it before myself so just curious as to how it works – feel free to point me to some docs that explain this stuff.

bjss-chris: Thanks. Leave it as is for now.

PR #436, opened 2021-10-11
Feature/tech debt more tests

MichaelRobinson-BJSS:
Add additional unit tests
Remove redundant code
Prevent commit of generated artefacts
Add npm install and gulp min to build pipeline


bjss-colin (dismissed):
Looks good to me

azure-pipelines.yml
```diff
@@ -60,31 +60,44 @@ jobs:
       nugetConfigPath: 'NuGet.config'
       arguments: '-v n'
 
+  - task: Npm@1
+    displayName: 'Run npm install'
+    inputs:
+      command: 'install'
+      workingDir: 'src/NHSD.GPIT.BuyingCatalogue.WebApp'
+
```
bjss-chris, line 68: I think we should add a task to update `npm` to the latest version before this task (all the old node repos and the old Identity repo have a similar task, if I remember correctly).

src/NHSD.GPIT.BuyingCatalogue.WebApp/gulpfile.js
```diff
@@ -2,12 +2,11 @@
...
 const gulp = require("gulp"),
     concat = require("gulp-concat"),
-    cssmin = require("gulp-cssmin"),
-    htmlmin = require("gulp-htmlmin"),
+    cssmin = require("gulp-cssmin"),   
     terser = require('gulp-terser'),
     merge = require("merge-stream"),
     del = require("del"),
-    sass = require("gulp-sass"),
+    sass = require("gulp-sass")(require('sass')),
```
bjss-chris, line 9: This looks odd. Is this additional `require('sass')` needed? How is it being used if it's not being assigned to anything?

src/NHSD.GPIT.BuyingCatalogue.WebApp/gulpfile.js
```diff
@@ -37,7 +36,7 @@ gulp.task("min:js", async function () {
     return merge(tasks);
 });
 
-gulp.task("min:css", function () {
+gulp.task("min:css", async function () {
```
bjss-chris, line 39: Do this need to be `async`? I can't see any `await` statements.

src/NHSD.GPIT.BuyingCatalogue.WebApp/package-lock.json
```diff
@@ -1,6632 +1,8 @@
 {
-  "name": "NHSD.GPIT.BuyingCatalogue.WebApp",
+  "name": "nhsd.gpit.buyingcatalogue.webapp",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
bjss-chris, line 4: Looks like you need to upgrade your `npm` version as the lock file version has been downgraded.

azure-pipelines.yml
MichaelRobinson-BJSS: Done. Added task to updated Node to 14.x (which includes npm) although MS keep the node version in the agent updated frequently. Can't see an obvious way to update npm on its own. I've basically copied the old repos.

src/NHSD.GPIT.BuyingCatalogue.WebApp/package-lock.json
MichaelRobinson-BJSS: Done

bjss-chris (approved):
I think we should add the task to update `npm` but we can do that in a separate PR.

PR #440, opened 2021-10-12
Add associated service statuses

PhysicsJon:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/AssociatedServices/EditAssociatedServiceModel.cs
```diff
@@ -16,5 +18,23 @@ public EditAssociatedServiceModel(CatalogueItem catalogueItem, CatalogueItem ass
...
+            if (!string.IsNullOrEmpty(AssociatedService.AssociatedService.Description)
+                && !string.IsNullOrEmpty(AssociatedService.AssociatedService.OrderGuidance)
+                && !string.IsNullOrEmpty(AssociatedService.Name))
+                return TaskProgress.Completed;
+
+            return TaskProgress.NotStarted;
+        }
+
+        public TaskProgress ListPriceStatus()
+        {
```
bjss-colin, line 33: Can be shortened to a ternary

bjss-colin (approved):
Approved. One suggested change but it's not mandatory.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/AssociatedServices/EditAssociatedServiceModel.cs
```diff
@@ -16,5 +18,23 @@ public EditAssociatedServiceModel(CatalogueItem catalogueItem, CatalogueItem ass
...
+            if (AssociatedService.CataloguePrices is not null && AssociatedService.CataloguePrices.Any())
```
bjss-chris, line 34: `CataloguePrices` should never be null (as it's initialized in the constructor). With the simpler expression, a ternary here as per @bjss-colin's suggestion would make sense.

PR #492, opened 2021-11-05
Add temporal tables

MichaelRobinson-BJSS:
Added stored procedures to enable or disable temporal tables
Enabled temporal tables for all IAudited tables


bjss-colin (approved):
Looks good to me

PR #493, opened 2021-11-08
Add, edit and delete service levels

bjss-colin:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/ServiceLevelAgreements/ServiceLevelCredits.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Admin.Models.ServiceLevelAgreements
+{
+    public enum ServiceLevelCredits
+    {
+        No = 0,
+        Yes = 1,
+    }
+}
```
PhysicsJon, line 8: Do we not have an extension that could do this and make the db column a bit?

PhysicsJon (changes requested):
A small change (look for BooleanExtensions)

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/ServiceLevelAgreements/ServiceLevelCredits.cs
bjss-colin: I initially added this for the frontend. Will take a further look to see if `bool` can instead be used for the radio list.

bjss-colin: Removed this and switched to using bool

PhysicsJon (approved):

PR #506, opened 2021-11-10
Remove unneeded "Go back" and "Save and Continue" content

leerogers-bjss:
Remove all unneeded setting of "Go back" LinkText

Remove all unneeded setting of "Save and Continue" Submit Button text.

Remove testing for "Go back" and "./" from unit tests that don't need them

Update RoadMapModel.cs, RoadMap.cshtml and controllers to set the go back url using Url.Action

Update RoadMapModel.cs to use constructors to set values instead of random functions.

Update FeaturesModel.cs, Features.cshtml and controllers to set the go back url using Url.Action.

Update FeatuesModel.cs to use constructor instead of random functions to set values

Update SolutionCheckEpicsModel.cs, Capabilities.cshtml, AdditionalCapabilities.cshtml and controllers to set go back url using Url.Action

Update tests for FeaturesModel.cs and RoadMapModel.cs to take into account the use of constructors.

### DIFF
--- a/src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/SolutionCheckEpicsModel.cs
+++ b/src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/SolutionCheckEpicsModel.cs
@@ -3,10 +3,11 @@ using System.Linq;
 using NHSD.GPIT.BuyingCatalogue.EntityFramework.Catalogue.Models;
 using NHSD.GPIT.BuyingCatalogue.EntityFramework.Ordering.Models;
 using NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Admin.Models;
+using NHSD.GPIT.BuyingCatalogue.WebApp.Models;
 
 namespace NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Solutions.Models
 {
-    public sealed class SolutionCheckEpicsModel : INoNavModel
+    public sealed class SolutionCheckEpicsModel : NavBaseModel, INoNavModel
     {
         public SolutionCheckEpicsModel()
         {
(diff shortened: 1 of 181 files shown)


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/SolutionCheckEpicsModel.cs
```diff
@@ -3,10 +3,11 @@
 using NHSD.GPIT.BuyingCatalogue.EntityFramework.Catalogue.Models;
 using NHSD.GPIT.BuyingCatalogue.EntityFramework.Ordering.Models;
 using NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Admin.Models;
+using NHSD.GPIT.BuyingCatalogue.WebApp.Models;
 
 namespace NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Solutions.Models
 {
-    public sealed class SolutionCheckEpicsModel : INoNavModel
+    public sealed class SolutionCheckEpicsModel : NavBaseModel, INoNavModel
```
bjss-colin, line 10: Should `INoNavModel` be removed? Otherwise it reads as being a nav model, but not being one at the same time.

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/SolutionCheckEpicsModel.cs
leerogers-bjss: yeah we need to keep it since INoNavModel has some weird fields that need to be set for the marketing page.

This wasn't named by myself.

PhysicsJon (approved):
Other than renaming the No Nav interface, it looks good

bjss-colin (approved):
As Jon said, we should probably rename `INoNavModel` interface at some point

PR #517, opened 2021-11-15
Use hyperlink if description is URL

bjss-colin:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/Description.cshtml
```diff
@@ -49,7 +49,15 @@
             {
                 <h2>Full description</h2>
 
-                <p style="white-space: pre-line">@Model.Description</p>
+                <p>
+                    @if(Model.Description.StartsWith("http://") || Model.Description.StartsWith("https://"))
```
PhysicsJon, line 53: This won't work with a description that is similar to `https://www.example.com/ this is a website`

PhysicsJon (changes requested):
We can't do this in the standard description text field as it could cause significant issues if the description is not purely a URL

bjss-colin:
Closing this and leaving as is for now. Had a discussion with @PhysicsJon and we both agreed that there are too many security implications with this change.

PR #521, opened 2021-11-15
Add URL Validator

bjss-colin:


src/NHSD.GPIT.BuyingCatalogue.Services/Validation/UrlValidator.cs
```diff
@@ -0,0 +1,47 @@
...
+        [SuppressMessage("Design", "CA1054:URI-like parameters should not be strings", Justification = "URL as string is intentional")]
+        public async Task<bool> IsValidUrl(string url)
+        {
+            if (string.IsNullOrWhiteSpace(url))
+                throw new ArgumentException("Url cannot be null, empty or whitespace", nameof(url));
+
+            var localUrl = url;
+            if (!localUrl.StartsWith("http://", StringComparison.CurrentCultureIgnoreCase)
+                && !localUrl.StartsWith("https://", StringComparison.CurrentCultureIgnoreCase))
+                localUrl = $"https://{localUrl}";
```
MichaelRobinson-BJSS, line 28: Shouldn't we force user to provide a fully qualified Url? If not fully qualified, how does this impact the link in public browse?

bjss-colin: This would have implications if the user specified a URL, without a protocol, where that website supports only HTTP. At the moment we don't have any content on the form inputs or validation errors that outline what format is expected.

leerogers-bjss: i agree we shouldn't force a protocol, but throw an error asking for one.

bjss-colin: Having a call with Luke @ 1430. Will raise the content question then.

PR #537, opened 2021-11-17
16477: Edit and Delete GP Connect Integrations

PhysicsJon:


bjss-colin (approved):
Can't see any issues in this. Looks good!

PR #542, opened 2021-11-18
Fix interoperability headings

bjss-colin:


PhysicsJon (approved):
I think we need to do a tidy up of the E2E interoperability tests as they're spread out a bit. We can do that after the release though

PR #557, opened 2021-11-24
Fix Filter Caching to only invalidate on solution update

leerogers-bjss:
Change Filter Caching Logic to only remove cached items when a solutions published status, details page, or Capabilities are edited.

Fix Typescript not actually working (no get request and broken function)

Add RemoveAll To FilterCache logic.

Add Updated Keys to appsettings.json

Update tests.


src/NHSD.GPIT.BuyingCatalogue.Services/Caching/FilterCache.cs
```diff
@@ -33,8 +44,18 @@ public void Remove(IEnumerable<string> filterKeys)
             }
         }
 
-        public void Set(string filterKey, string content, DateTime expiration)
-            => memoryCache.Set(GetCacheKey(filterKey), content, expiration);
+        public void RemoveAll()
+        {
+            var cancellationToken = memoryCache.Get<CancellationTokenSource>(filterCacheKeySettings.CancellationSourceKey);
```
bjss-colin, line 49: Could you use the `CancellationToken` property here?

bjss-colin (changes requested):
Only one thing spotted, rest looks good.

src/NHSD.GPIT.BuyingCatalogue.Services/Caching/FilterCache.cs
leerogers-bjss: Theres a difference between CancellationTokenSource and CancelltationToken, so need to keep the full object definition

bjss-colin (approved):

PR #576, opened 2021-12-02
Fix StringLength values to match database schema

bjss-colin:


MichaelRobinson-BJSS (dismissed):

leerogers-bjss (changes requested):
need to update the sql as requested

leerogers-bjss (approved):

MichaelRobinson-BJSS (approved):

PR #577, opened 2021-12-02
Fix performance issue with solution loading

PhysicsJon:
So this got big
- [ ] Add status model
- [ ] Update controllers and services to use status model
- [ ] Update tests to cover changes
- [ ] Remove status methods from models that no longer require them


bjss-colin (approved):
Excellent work

PR #597, opened 2021-12-08
Feature/get order performance

MichaelRobinson-BJSS:


PhysicsJon (approved):
Seems good to me, but do we need to add null checks for the calls to GetOrderThin?

leerogers-bjss (approved):

PR #616, opened 2022-01-10
Add LastPublished and update ordering

bjss-colin:


PhysicsJon (approved):
I'm approving with the assumption that some form of E2E tests are going to be added in a future PR

PR #620, opened 2022-01-13
Add unit test framework for view components

PhysicsJon:


bjss-colin (approved):
That was easy, nice one!

PR #621, opened 2022-01-13
Implement Notify platform to replace SMTP handling

bjss-colin:


PhysicsJon (approved):
Seems pretty handy to me

PR #633, opened 2022-01-20
Redirect to Supplier page when supplier already exists

bjss-colin:


leerogers-bjss (changes requested):
The Acceptance criteria on the bug is wrong (see my comment on the bug)
They should be redirected to the Supplier endpoint in the SupplierController not back to the order task list

leerogers-bjss (approved):

PR #656, opened 2022-01-31
Add autocomplete filter to solutions browsing page

leerogers-bjss:
Update autocomplete design and functionality to follow a more GDS like specification

Add SuggestionEndpoint to SolutionsController

Add GetSolutionsBySearchTerm to SolutionsFilterService

Change autocompleteConfig to be agnostic to the number of items on the page.

update tsconfig.json to exlude node_modules explicitly.

update bundleconfig.json to minify the autocompleteConfig file


bjss-colin (approved):
Approved with one suggested change.

Also, the min.js & .map files shouldn't need to be version controlled as they're build-time artifacts. Although, given how little js files we have I don't think it's a big deal.

PR #700, opened 2022-02-23
Import practice lists

magnusjohn-bjss:


terraform/modules/bc_webapp_pri/variables.tf
```diff
@@ -46,6 +46,10 @@ variable "auth_pwd" {
   type      = string
   sensitive = true
 }
+variable "hangfire_pwd" {
```
steves-bjss: This will need the password generation code adding too.  I will ping you about it separately.

steves-bjss (changes requested):

steves-bjss (approved):

Okechukwu802 (approved):

PR #701, opened 2022-02-24
Changed odsCode to internalOrgId where appropriate in the service layer

MichaelRobinson-BJSS:


src/NHSD.GPIT.BuyingCatalogue.Services/Session/OrderSessionService.cs
```diff
@@ -119,12 +119,12 @@ public async Task<CreateOrderItemModel> InitialiseStateForEdit(string odsCode, C
 
                 if (state.CatalogueItemType == CatalogueItemType.AssociatedService)
                 {
-                    var organisation = await organisationService.GetOrganisationByInternalIdentifier(odsCode);
-                    state.ServiceRecipients = new List<OrderItemRecipientModel> { new() { OdsCode = odsCode, Name = organisation.Name } };
+                    var organisation = await organisationService.GetOrganisationByInternalIdentifier(internalOrgId);
+                    state.ServiceRecipients = new List<OrderItemRecipientModel> { new() { OdsCode = internalOrgId, Name = organisation.Name } };
```
magnusjohn-bjss: Looks like a missed OdsCode model property here

MichaelRobinson-BJSS: Good spot, although not quite that simple. The actual orders still use ODS codes and will continue to do so until an equivalent for ASC is determined. The above code should be using the organisations ExternalIdentifier rather than the internalOrgId.

magnusjohn-bjss (approved):

PR #705, opened 2022-02-25
Post order guidance

magnusjohn-bjss:


src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -228,6 +235,20 @@ public async Task CompleteOrder(CallOffId callOffId, string internalOrgId)
                 templateId,
                 personalisation);
 
+            var userEmail = dbContext.Users.Single(x => x.Id == userId).Email;
+            var pdfData = pdfService.Convert(orderSummaryUri);
```
ghost, line 234: Given you're doing some PDF stuff here, it's possibly worth removing the previous `await` on the `emailService` to let it run asynchronously. Instead capture the `Task` in a variable and await it further down in a `Task.WhenAll` when you're sending the PDF email. That way this request won't go on for too long.

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
magnusjohn-bjss: Wrapped the three async calls in a Task.WhenAll that runs post-pdf generation

bjss-colin (approved):

Okechukwu802 (approved):

PR #755, opened 2022-03-29
Create DbUp project

bjss-colin:
Adds a new project that uses DbUp which allows for automated deployment of SQL artifacts produced during a sprint.

I've included an initial script which does nothing just to kick-start the naming convention.


bjss-colin:
Closing for now, to be picked up again in the future.

PR #778, opened 2022-05-04
Confirm prices

magnusjohn-bjss:


bjss-colin (approved):
Looks good to me.

Although, if we're using `SolutionListPriceService` more generically then we should probably rename this to be generic or move the methods into the existing `ListPriceService`, but this is a problem for future us to worry about.

The idea behind keeping it separate initially was so Additional and Associated services could have their own services defined that accepted the Service Id and the Solution Id which those services belong to. Then again, this could be a series of overloaded methods on the interface.

steves-bjss (approved):

steves-bjss:
Good test coverage of acceptance criteria.

PR #792, opened 2022-05-16
Quantity selection & routing

magnusjohn-bjss:


Okechukwu802 (approved):

src/NHSD.GPIT.BuyingCatalogue.Services/Routing/Providers/SelectServiceRecipientsProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                    RouteValues = new
+                    {
+                        routeValues.InternalOrgId,
+                        routeValues.CallOffId,
+                        routeValues.CatalogueItemId,
+                    },
+                };
+            }
+
+            // TODO: Error page if no published prices?
```
ghost, line 59: Error page and logging an exception/error sounds wise. We do have guards around the published status of a solution so there should always be a list price associated with an item. I'd only expect this to be hit where some database changes have occurred that corrupt the data, hence why logging is needed so we can figure out what went wrong here.

bjss-colin (approved):

PR #808, opened 2022-06-06
Bump Nuget Versions, Fix Validators and Delete UI Component E2E Tests

leerogers-bjss:
Update all nuget packages to the latest version

Delete the UI.Components.E2E tests project.

Update all Validators to remove any asynchronous functions and make the synchronous (requires for fluentvalidation v11)
Remove all calls to .Casecade(Behaviour.Stop) because it is now being defined at a global level

Change IWebHost to IHost in Integration tests due to IWebHost initialization being depreciated in .net core 3.0.


steves-bjss (changes requested):
The xUnit versions appear to be mismatched across the projects - some are 2.4.1 and others are 2.4.2-pre.12.  

Could we make them consistant please and also confirm if we want to upgrade to a pre-release version of xUnit.  Thanks!

steves-bjss (approved):

PR #821, opened 2022-06-22
Review order pdf

magnusjohn-bjss:


bjss-colin (approved):
Looks good, nice simplification of the calculation logic.

steves-bjss (approved):

PR #830, opened 2022-06-30
add estimation period for quantity

leerogers-bjss:
Rename estimation period on order item price to billing period.

update order form quantity pages to display different content depending on the estimation period.

Add RangeDefinition logic to flat price tests.

Add SetOrderItemEstimation Tests


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Views/Shared/ListPrices/AddEditFlatListPrice.cshtml
```diff
@@ -121,6 +121,11 @@
                 asp-for="UnitDefinition"
                 character-count="false" />
 
+            <nhs-input
+                label-text="Range definition"
```
ghost: Range definition isn't the correct term here for the label text as a flat price has no ranges. Might be worth running this by Luke to coin up a better term.

bjss-colin (changes requested):

magnusjohn-bjss (approved):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Views/Shared/ListPrices/AddEditFlatListPrice.cshtml
magnusjohn-bjss: Not sure I agree with this, range definition is the term we use for the plural of the unit description (i.e. "beds" when the unit is "per bed"), it has descriptive lede text to make that clear and should be consistent between flat & tiered prices.

bjss-colin (approved):
Approved for now but it's worth re-reviewing the `Range definition` term. I still don't think it's the right term here.

bjss-colin (approved):

PR #837, opened 2022-07-06
Fix for unpublished additional services on order pathway

magnusjohn-bjss:


src/NHSD.GPIT.BuyingCatalogue.Services/AdditionalServices/AdditionalServicesService.cs
```diff
@@ -73,14 +73,18 @@ public async Task<CatalogueItem> GetAdditionalServiceWithCapabilities(CatalogueI
...
             if (!catalogueItemId.HasValue)
             {
                 return new List<CatalogueItem>();
             }
 
-            return await BaseQuery(catalogueItemId.Value).ToListAsync();
+            var output = await BaseQuery(catalogueItemId.Value).ToListAsync();
+
+            return publishedOnly
+                ? output.Where(x => x.PublishedStatus == PublicationStatus.Published).ToList()
```
ghost: Append this to the `IQueryable` instead prior to calling `ToListAsync` on `BaseQuery`. That way it's evaluated at SQL time rather than being done in-memory

bjss-colin (approved):
One suggested improvement, but it's not a massive issue so also approved. Up to you whether you want to change it.

leerogers-bjss (approved):

PR #839, opened 2022-07-07
Only use the practice list size for Per Patient

bjss-colin:
Removes the additional page which was previous added for Per Service Recipient.

Shifts the quantity logic from the controller to the model for Per Service Recipient.

Hardcode the column heading to "Quantity" as using the Units or Range Description doesn't read or format very well.


magnusjohn-bjss (approved):
Nice, much better without the second per service recipient view imo

PR #840, opened 2022-07-07
Reset order item quantities on price change

magnusjohn-bjss:


bjss-colin (approved):
Good call with checking the price quantity basis. In a Per Patient => Per Patient scenario it makes sense not to reset the quantities.

PR #861, opened 2022-07-25
Implementation plan & milestones

magnusjohn-bjss:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Controllers/Contracts/ImplementationPlanController.cs
```diff
@@ -0,0 +1,134 @@
...
+            this.implementationPlanService = implementationPlanService ?? throw new ArgumentNullException(nameof(implementationPlanService));
+            this.orderService = orderService ?? throw new ArgumentNullException(nameof(orderService));
+            this.solutionsService = solutionsService ?? throw new ArgumentNullException(nameof(solutionsService));
+        }
+
+        [HttpGet("default")]
+        public async Task<IActionResult> DefaultImplementationPlan(string internalOrgId, CallOffId callOffId)
+        {
+            return View(
+                "~/Areas/Order/Views/Contracts/DefaultImplementationPlan.cshtml",
```
ghost, line 40: Shouldn't need the full path here, `DefaultImplementationPlan` by itself should work. It just depends on how verbose you want to be.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Views/AssociatedServices/AddAssociatedServices.cshtml
```diff
@@ -12,7 +12,7 @@
         <nhs-page-title title="@ViewBag.Title"
                         caption="Order @Model.CallOffId"
                         advice="These are products that support the implementation of a solution and can be bought independently. For example, training or data migration or arranging the merging or splitting of existing practices."/>
-
+
```
ghost: Can be removed

bjss-colin (approved):
Approved with 2 minor comments

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Controllers/Contracts/ImplementationPlanController.cs
magnusjohn-bjss: It shouldn't need the full path here, as both view & method have the same (unique) name, but leaving it out makes razor throw a fit in development view so I left it in. Can't see what the problem is but it affects both the views in this file, hence the full path specifications.

PR #871, opened 2022-08-04
Supplier contact validation

magnusjohn-bjss:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Validators/EditContactModelValidator.cs
```diff
@@ -8,51 +8,74 @@ namespace NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Admin.Validators
 {
     public sealed class EditContactModelValidator : AbstractValidator<EditContactModel>
```
ghost, line 10: Although `NewContactModel` and `EditContactModel` exist in two different areas of the system they both relate to the same entity, a supplier contact. `EditContactModel` could inherit the former, extending with its own properties (I.E: SolutionsReferencing), since they contain very similar logic and inputs.

Then the `NewContactModelValidator` could be included here with `Include(new NewContactModelValidator())` and `EditContactModelValidator` then performs a duplication check on top of the existing logic.

bjss-colin (dismissed):

bjss-colin (dismissed):

bjss-colin (approved):

PR #872, opened 2022-08-04
Clean up duplicated integration tests

kelly-bjss:


leerogers-bjss:
branch needs rebuilding

PR #874, opened 2022-08-05
Update Delete order page and button

leerogers-bjss:
Update secondary button to have a type or Primary, Secondary or Delete which defines it's colour.

Update all references to secondary button to use this new format

Update Delete order Page

Add / Update Delete order page tests

### DIFF
--- a/src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Controllers/DeleteOrderController.cs
+++ b/src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Controllers/DeleteOrderController.cs
@@ -40,6 +40,17 @@ namespace NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Order.Controllers
         [HttpPost]
         public async Task<IActionResult> DeleteOrder(string internalOrgId, CallOffId callOffId, DeleteOrderModel model)
         {
+            if (!ModelState.IsValid)
+                return View(model);
+
+            if (!model.SelectedOption.Value)
+            {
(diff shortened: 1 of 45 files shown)


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Controllers/DeleteOrderController.cs
```diff
@@ -40,6 +40,17 @@ public async Task<IActionResult> DeleteOrder(string internalOrgId, CallOffId cal
         [HttpPost]
         public async Task<IActionResult> DeleteOrder(string internalOrgId, CallOffId callOffId, DeleteOrderModel model)
         {
+            if (!ModelState.IsValid)
```
ghost, line 51: ~~Looks like the story acceptance criteria wasn't updated to include the following, but there also needs to be a check in the GET method to make sure the buyer can't delete completed orders by manually navigating to the delete URL with the order ID.~~

On second thought, this is probably best as a separate story that includes the navigation logic.

bjss-colin (changes requested):

bjss-colin (dismissed):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Controllers/DeleteOrderController.cs
leerogers-bjss: fixed in latest commit

bjss-colin (approved):

PR #875, opened 2022-08-08
Update practice list sizes

magnusjohn-bjss:


bjss-colin (approved):
Looks good, nice trick with the `IsRelational` check.

PR #893, opened 2022-08-31
Update NHSD.GPITBuyingCatalogue.Database.sqlproj

leerogers-bjss:
set InsertOrderTriageValues.sql to not build since it's a post deployment script and shouldn't build with the application


steves-bjss (approved):

leerogers-bjss:
closing because this bug is fixed in another PR

PR #899, opened 2022-09-07
Role based access pt. 2

bjss-colin:


magnusjohn-bjss (approved):
Some potential future issues when multiple roles are implemented with .First() being used in a couple of places to determine a users' role

bjss-colin:
> Some potential future issues when multiple roles are implemented with .First() being used in a couple of places to determine a users' role

I thought about this too. On second thought, I don't think we'll need multiple roles any time soon as, when account managers are implemented, we can update the Buyer _policy_ to grant access Buyer and Account Manager, with a separate Account Manager policy also existing alongside the existing two.

Although, if we do go the route of having multiple roles where a user is an account manager and a buyer, the current code orders the roles by name so Account Manager appears first for the user management page to choose the correct role in the radio list.

imranbashir01 (approved):

PR #907, opened 2022-09-16
Terraform changes for Disaster Recovery

bjss-colin:
Currently a draft raised for awareness.

* Updates the AzureRM provider version
* Retrieves the KeyVault secrets, from the primary environment, when the current environment is DR
* Moves KeyVault to a module that is conditionally deployed when the environment is not DR **
* Conditionally deploys a SQL instance and the databases when the environment is not DR

The KeyVault module transition is a destructive change and will require `terraform mv` to be run to move the existing KeyVault state binding to the new name of `module.azurerm_key_vault.keyvault`


terraform/resource_groups.tf
```diff
@@ -16,23 +16,6 @@ resource "azurerm_resource_group" "storageaccount" {
   }
 }
 
-resource "azurerm_resource_group" "keyvault" {
```
ghost, line 19: Moved to the KeyVault module main.tf

terraform/resource_groups.tf
```diff
@@ -16,23 +16,6 @@ resource "azurerm_resource_group" "storageaccount" {
...
-  name           = "${var.project}-${var.environment}-rg-keyvault"
-  location       = var.region
-  tags = {
-    environment  = var.environment,
-    architecture = "new"
-  }
-}
-
-resource "azurerm_resource_group" "sql-server" {
```
ghost, line 28: Moved to the bc_sql_server module main.tf

terraform/modules/keyvault/keyvault_access_policy.tf
```diff
@@ -0,0 +1,169 @@
...
+    "Purge",
+    "Recover",
+    "Restore",
+    "SetIssuers",
+    "Update"
+
+  ]
+}
+
+# Should we be managing a resource that we don't provision or deploy?
```
bjss-colin, line 115: Commented this out with a reason. This changes properties on a KeyVault instance that isn't provisioned by this terraform configuration. We should either bring the core KeyVault instance under this tf config or maybe scrap this block.

azure-pipelines.yml
```diff
@@ -31,163 +31,163 @@ jobs:
...
-      $id=docker images --filter "label=dacpac=true" -q | Select-Object -First 1
-      docker create --name dacpaccontainer $id      
-      mkdir $(Build.ArtifactStagingDirectory)/dacpacs
-      docker cp dacpaccontainer:/deploy-db/. $(Build.ArtifactStagingDirectory)/dacpacs
-      docker rm dacpaccontainer
-    displayName: 'Copy DACPACs'
-
-  - publish: $(build.artifactStagingDirectory)/dacpacs
-    artifact: dacpacs
+# - job: dockerBuildAndPush
```
steves-bjss: Can you re-enable the rest of the build please.

steves-bjss (changes requested):

steves-bjss (approved):

PR #928, opened 2022-10-12
Import service recipients

bjss-colin:


imranbashir01 (approved):

src/NHSD.GPIT.BuyingCatalogue.Services/Csv/CsvService.cs
```diff
@@ -129,27 +129,6 @@ public async Task<int> CreatePatientNumberCsvAsync(int orderId, MemoryStream str
 
         private static string TimeUnitDescription(TimeUnit? timeUnit) => timeUnit?.Description() ?? string.Empty;
```
dssherlock, line 131: Why move this into a base class? CsvClass sounds likes it's pretty generic anyway? Couldn't the ServiceRecipientImportService just inherit that instead? I haven't seen the whole file so it may be very obvious!

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Controllers/SolutionSelection/ImportServiceRecipientsController.cs
```diff
@@ -0,0 +1,342 @@
...
+            });
+    }
+
+    [HttpGet("download-template")]
+    public async Task<IActionResult> DownloadTemplate(
+        string internalOrgId,
+        CallOffId callOffId,
+        CatalogueItemId catalogueItemId,
+        ServiceRecipientImportMode? importMode = ServiceRecipientImportMode.Edit)
+    {
```
dssherlock, line 228: I'm not familiar with this syntax?

dssherlock (approved):

src/NHSD.GPIT.BuyingCatalogue.Services/Csv/CsvService.cs
ghost: `WriteAsync` is the generic method that was moved to a base class as it's being reused in the original `CsvService`, which generates order CSVs, and the new Service Recipient service that handles the importing of service recipients from a CSV.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Order/Controllers/SolutionSelection/ImportServiceRecipientsController.cs
ghost: `_` is the discard operator which indicates to the linter that it should ignore unused variables.

The parameters are part of the `Route` applied to the controller since those params are common across all methods, except this one.

Doing it this way prevents the linter from raising a warning about unused parameters. Conversely, removing the parameters from the method signature would cause a warning on the `Route` attribute since not all URL parameters would be passed into every method.

PR #948, opened 2022-11-11
Edit User Accounts

dssherlock:


imranbashir01 (dismissed):

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/AccountManagement/EditUser.cs
```diff
@@ -0,0 +1,184 @@
...
+using NHSD.GPIT.BuyingCatalogue.EntityFramework.Users.Models;
+using NHSD.GPIT.BuyingCatalogue.Framework.Extensions;
+using NHSD.GPIT.BuyingCatalogue.ServiceContracts.Identity;
+using NHSD.GPIT.BuyingCatalogue.WebApp.Areas.AccountManagement.Controllers;
+using Polly;
+using Xunit;
+
+namespace NHSD.GPIT.BuyingCatalogue.E2ETests.Areas.AccountManagement
+{
+    public sealed class EditUser : AccountManagerTestBase, IClassFixture<LocalWebApplicationFactory>
```
ghost: Let's implement `IDisposable` here to clean up any changes that the tests make. Tests should ideally clean up after themselves. As I'm currently in the process of improving the test times by using a shared database instance, it becomes more obvious that tests aren't resetting state back to what it was prior to changes, and it has a knock-on effect to other tests

For instance, `EditUser_EditUser_ExpectedResult` will result in User ID 3 (Bob, admin account) becoming an Account Manager and also Inactive.

The original object doesn't have to be cloned, instead the `Dispose` method can simply set the Role back to Admin and Inactive back to Active.

bjss-colin (changes requested):

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Areas/AccountManagement/EditUser.cs
dssherlock: Yep sure! Tbh I thought the tests already had a dispose method. I'll take a look

bjss-colin (approved):

PR #960, opened 2022-11-25
Consume TRUD data

bjss-colin:
Implemented with the same contract and logic as the existing `OdsService` for backwards compatibility.

The plan, in the future, is to remove the tuple return type and perform some cleanup.


src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Organisations/IOdsService.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
```
dssherlock: Is this needed if nothing else in the file has changed?

src/NHSD.GPIT.BuyingCatalogue.WebApp/appsettings.json
```diff
@@ -62,7 +62,8 @@
     "ApiBaseUrl": "https://directory.spineservices.nhs.uk/ORD/2-0-0",
     "BuyerOrganisationRoleIds": [ "RO98", "RO177", "RO213", "RO272" ],
     "GpPracticeRoleId": "RO177",
```
dssherlock, line 64: Not sure if it matters but everything else in this struct is Pascal Case?

dssherlock (dismissed):

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Organisations/IOdsService.cs
ghost: Good spot, can't remember what my thinking/planning was around this.

src/NHSD.GPIT.BuyingCatalogue.WebApp/appsettings.json
ghost: At some point we need to update configs and how we handle environment variables as there's definitely room for improvement.

Semantically speaking, and although there are no rules set in stone, pascal case is incorrect for JSON. Camel and snake case are the recommended naming conventions for the sake of consistency.

dssherlock: Yep fair enough-i cant say i have a strong opinion on it!!

dssherlock (approved):

imranbashir01 (approved):

PR #961, opened 2022-11-25
Edit User-Max 2 Account Managers

dssherlock:
Edit E2E tests still to do


imranbashir01 (dismissed):

imranbashir01 (dismissed):

src/NHSD.GPIT.BuyingCatalogue.Services/Users/UsersService.cs
```diff
@@ -150,14 +150,15 @@ public async Task<bool> EmailAddressExists(string emailAddress, int userId = 0)
             return user != null;
         }
 
-        public async Task<bool> IsAccountManagerLimit(int organisationId)
+        public async Task<bool> IsAccountManagerLimit(int organisationId, int userId = 0)
```
ghost, line 153: It's probably best to use a nullable int here as there's no guarantee a user ID won't be 0

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.Services/Users/UsersService.cs
dssherlock: This is how the function checking if email address exists (see line above) works so thought this was a fair assumption?

src/NHSD.GPIT.BuyingCatalogue.WebApp/Validation/Organisation/UserDetailsModelValidator.cs
```diff
@@ -33,13 +35,20 @@ public UserDetailsModelValidator(IUsersService usersService)
...
                 .NotEmpty()
                 .WithMessage("Select an account status");
         }
 
         private bool NotBeDuplicateUserEmail(string emailAddress, int userId) => !usersService.EmailAddressExists(emailAddress, userId).GetAwaiter().GetResult();
+
+        private bool AccountManagerLimit(string accountType, int orgId, int userId, bool disabled)
+        {
+            return !usersService.IsAccountManagerLimit(orgId, userId).Result
+                || accountType != OrganisationFunction.AccountManager.Name || disabled;
```
ghost: Put `accountType != OrganisationFunction.AccountManager.Name || disabled` at the start of this check to prevent a database hit when buyers are being added

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Validation/Organisation/UserDetailsModelValidator.cs
dssherlock: Yep good call

src/NHSD.GPIT.BuyingCatalogue.Services/Users/UsersService.cs
ghost: Hadn't noticed that one before, even though I changed a line in the code some months back.

Having checked the schema I can see the ID column is an `IDENTITY(1,1)` so it should be alright for now. We can change in the future if it becomes an issue.

bjss-colin (approved):

PR #966, opened 2022-11-30
Ordering party contact amendment status

bjss-colin:


src/NHSD.GPIT.BuyingCatalogue.Framework/Comparers/Ordering/ContactEqualityComparer.cs
```diff
@@ -0,0 +1,29 @@
...
+        if (ReferenceEquals(x, y)) return true;
+        if (ReferenceEquals(x, null)) return false;
+        if (ReferenceEquals(y, null)) return false;
+        if (x.GetType() != y.GetType()) return false;
+
+        return x.FirstName == y.FirstName
+            && x.LastName == y.LastName
+            && x.Email == y.Email
+            && x.Phone == y.Phone
+            && x.LastUpdated == y.LastUpdated;
```
magnusjohn-bjss: Do we want last updated in the comparison? Two otherwise identical records won't be considered equal if different users save them, which doesn't seem right

magnusjohn-bjss (dismissed):

magnusjohn-bjss (approved):

imranbashir01 (approved):

PR #970, opened 2022-12-02
US: 21447 -- Removed leading zero from date by chaning the date format.

DeepikaDK001:


magnusjohn-bjss (commented):
Looking good so far, but there are a number of other references to the date format outside of views. If you do a solution-wide search for `dd MMMM yyyy` and `.ToLongDateString()` you should be able to see the others

PR #974, opened 2022-12-06
US-21447 change in Date format for Date Created in the Manage Orders

DeepikaDK001:


magnusjohn-bjss (dismissed):

imranbashir01 (dismissed):

bjss-colin (changes requested):
If the admin pages are also being changed then it might be worth updating all date references in that case.

Doing a search for dd MM and ToLongDateString brings back a few results. The only place where the date format _**must**_ remain unchanged is `GpPracticeService`

imranbashir01 (dismissed):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Properties/launchSettings.json
```diff
@@ -1,54 +0,0 @@
-{
```
ghost: Looks like there's just one last test that's failing, but we also need to keep this file as without it we can't run the app locally

bjss-colin (changes requested):

imranbashir01 (dismissed):

bjss-colin (approved):

imranbashir01 (approved):

PR #981, opened 2022-12-13
Max two account managers and display account type in User list

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.Framework/Extensions/AspNetUserExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static string GetRoleName(this AspNetUser user)
         public static string GetDisplayRoleName(this AspNetUser user)
         {
             var roleName = GetRoleName(user);
-            return OrganisationFunction.FromName(roleName)!.DisplayName;
+            return roleName is not null ? OrganisationFunction.FromName(roleName)!.DisplayName : string.Empty;
```
ghost: Assuming you've hit the same problem, the agency user is what causes this to fail.

Considering Role Name should never be empty, I'd say it's an incorrect behavior to return an empty string. Since we don't use the agency user I've disabled this in the docker-compose file.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Validators/Users/UserDetailsModelValidator.cs
```diff
@@ -74,12 +74,10 @@ private static bool BelongToCorrectOrganisation(string accountType, string selec
...
 
-        private bool AccountManagerLimit(string accountType, string selectedOrganisationId, int userId)
+        private bool AccountManagerLimit(string accountType, string selectedOrganisationId, int userId, bool disabled)
         {
-            if (accountType != OrganisationFunction.AccountManager.Name)
-                return true;
-
-            return !usersService.IsAccountManagerLimit(int.Parse(selectedOrganisationId), userId).Result;
+            return accountType != OrganisationFunction.AccountManager.Name || disabled
+                || !usersService.IsAccountManagerLimit(int.Parse(selectedOrganisationId), userId).Result;
```
ghost: Rather than `int.Parse`, the model property `SelectedOrganisationId` could become an `int?` to leave ASP.NET Core handle the model binding.

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.Framework/Extensions/AspNetUserExtensions.cs
dssherlock: Yep ok fair enough. It was also a safety check but providing a user always has a role assigned then thats fine.

bjss-colin (approved):

PR #989, opened 2022-12-16
Gulp.js File to update NHSFrontend and GovUk in wwwroot

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.WebApp/gulpfile.js
```diff
@@ -61,4 +79,4 @@ function getBundles(regexPattern) {
     });
 }
 
-gulp.task("min", gulp.series("clean", "sass", "min:js", "min:css"));
+gulp.task("min", gulp.series("clean", "sass", "min:js", "min:css", "nhsuk", "govuk"));
```
ghost: If you reorder these steps, to put these after the `sass` task, and add the GOV UK `all.js` file into the bundleconfig.json, in Web App root, then you should be able to piggy-back on the existing minification process in the `min:js` gulp task. Uglify shouldn't needed at that point since the `govuk` and `nhsuk` tasks at that point will only be copying files from A to B.

I haven't tested this, but in theory it _should_ work

bjss-colin (dismissed):
Other than one suggestion, which isn't mandatory, this is a really good effort & approach.

src/NHSD.GPIT.BuyingCatalogue.WebApp/gulpfile.js
dssherlock: Ok cool I'll give it a try

bjss-colin (approved):

imranbashir01 (approved):

PR #992, opened 2022-12-19
Edit user inset text change

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/OrganisationBase/UserDetails.cshtml
```diff
@@ -23,7 +23,7 @@
                 <input type="hidden" asp-for="UserId" />
                 <input type="hidden" asp-for="BackLink" />
                 <input type="hidden" asp-for="IsDefaultAccountType" />
-                <input type="hidden" asp-for="MaxNumberOfAccountManagers" />
+                <input type="hidden" asp-for="MaximumAccountManagerMessage" />
```
ghost: `MaximumAccountManagerMessage` is a getter only so we shouldn't need a hidden input here as it uses the `UserId` to return a message.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/OrganisationModels/UserDetailsModel.cs
```diff
@@ -46,7 +51,13 @@ public string Title
 
         public int UserId { get; set; }
 
-        public int MaxNumberOfAccountManagers { get; set; }
+        public string MaximumAccountManagerMessage
+        {
+            get
+            {
+                return UserId == 0 ? string.Format(AddMaximumAccountManagerMessage, maxAccountManagers) : string.Format(EditMaximumAccountManagerMessage, maxAccountManagers);
```
ghost: Since `maxAccountManagers` is a private read-only field, I'd expect this to be the default value for integers, `0`, if an admin or user removes causes a form validation error, causing the `AddMaximumAccountManagerMessage` to appear instead of `EditMaximumAccountManagerMessage`

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/OrganisationBase/UserDetails.cshtml
dssherlock: Yep u r right this has changed so can be removed

bjss-colin (dismissed):

steves-bjss (approved):

PR #1006, opened 2023-01-15
Adding & removing additional services from an amended order

magnusjohn-bjss:


imranbashir01 (dismissed):

imranbashir01 (dismissed):

imranbashir01 (approved):

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Ordering/Models/Order.Custom.cs
```diff
@@ -208,5 +209,25 @@ public override int GetHashCode()
         {
             return Id.GetHashCode();
         }
+
+        public Order Clone()
```
magnusjohn-bjss, line 213: Using Newtonsoft rather than a faster json serialiser here as
a) the cyclic references used by ef (i.e. OrderItem -> OrderItemPrice -> OrderItem etc.) aren't supported by all serialisers
b) we've got interfaces included as properties on some classes (thanks lee!) which require custom type name handling to deserialise

bjss-colin (approved):

PR #1016, opened 2023-01-19
Unit tests

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/SolutionsController.cs
```diff
@@ -165,14 +165,14 @@ public async Task<IActionResult> Capabilities(CatalogueItemId solutionId)
             CatalogueItemId additionalServiceId)
         {
             var solution = await solutionsService.GetSolutionThin(solutionId);
-            if (solution is null)
+            if (solution is null || solution.Solution is null)
```
magnusjohn-bjss: These can be coalesced into `if (solution?.Solution is null)`

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/SolutionsController.cs
```diff
@@ -165,14 +165,14 @@ public async Task<IActionResult> Capabilities(CatalogueItemId solutionId)
...
                 return BadRequest($"No Solution found for Id: {solutionId}");
 
             if (solution.PublishedStatus == PublicationStatus.Suspended)
                 return RedirectToAction(nameof(Description), new { solutionId });
 
             var item = await additionalServicesService.GetAdditionalServiceWithCapabilities(additionalServiceId);
-            if (item is null)
+            if (item is null || item.AdditionalService is null)
```
magnusjohn-bjss: Ditto

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/SolutionDisplayBaseModel.cs
```diff
@@ -28,6 +28,9 @@ protected SolutionDisplayBaseModel()
             if (catalogueItem is null)
                 throw new ArgumentNullException(nameof(catalogueItem));
 
+            if (catalogueItem.Solution is null)
```
magnusjohn-bjss: These two checks can be coalesced into one (`if (catalogueItem?.Solution is null)`), it's generally not good practice to call out a property of an argument instead of the argument itself.

dssherlock: Yeah this was a bit of a compromise really. Ideally the constructor would be changed to take a Solution rather than a Catalogue item and then check for null but this is used literally all over the place and would result in massive amounts of refactoring.

magnusjohn-bjss (dismissed):

imranbashir01 (dismissed):

magnusjohn-bjss (approved):

PR #1052, opened 2023-02-23
Fix for validation for empty input does not trigger.

DeepikaDK001:


bjss-colin:
Please ensure PRs are raised _**only**_ when they are dev complete with all tests updated accordingly. Updating unit tests is part of the development process and not a review time activity.

PR #1069, opened 2023-03-17
Feature/21922 Create ICBs

dssherlock:


database/NHSD.GPITBuyingCatalogue.Database/PostDeployment/CreateIntegratedCareBoards.sql
```diff
@@ -0,0 +1,52 @@
+﻿DECLARE @icbRoleId AS nchar(5) = 'RO318';
+
+IF UPPER('$(SEED_ORGANISATIONS)') = 'TRUE' AND NOT EXISTS (SELECT * FROM organisations.Organisations WHERE OrganisationRoleId = @icbRoleId)
```
ghost: Let's change this so it's using the TEST_DATA argument rather than Seed Organisations as that would add the ICBs on preprod and production also. For the upper environments, this should be handled by the migration script.

It's also potentially worth using a MERGE statement for this as it may cause errors if the table is in a partial state.

database/NHSD.GPITBuyingCatalogue.Database/OdsOrganisations/Functions/GetParentICB.sql
```diff
@@ -0,0 +1,14 @@
+﻿CREATE FUNCTION ods_organisations.GetParentICB (@orderingPartyId int)
```
ghost: Let's remove this one for now as I think our approach has changed in terms of how we're navigating the structure

bjss-colin (changes requested):
2 minor suggestions

database/NHSD.GPITBuyingCatalogue.Database/OdsOrganisations/Functions/GetParentICB.sql
dssherlock: Im not sure how u mean?

This function is used through-out the migration script to obtain the Id of the newly created ICB organisations based on the Id of the CCG.

PR #1084, opened 2023-04-12
Feature/21941 gp local funding

dssherlock:
Limit the funding sources for GP Practices buyers so that they can only procure using local funding


src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderItemService.cs
```diff
@@ -161,7 +162,7 @@ public async Task DetectChangesInFundingAndDelete(CallOffId callOffId, string in
 
             if (item.TotalCost() == 0)
                 newFundingType = OrderItemFundingType.NoFundingRequired;
-            else if (item.Order.SelectedFramework.LocalFundingOnly)
+            else if (item.Order.SelectedFramework.LocalFundingOnly || item.Order.OrderingParty.OrganisationType == OrganisationType.GP)
```
ghost: This logic is duplicated in quite a lot of places. Let's split this out to a method on the order object instead. See `IsPerServiceRecipient` on `OrderItemPrice` for an example of a similar setup

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderItemService.cs
dssherlock: Yep i agree-wanted to get it in and working but yes needs refactoring.

bjss-colin (approved):

PR #1101, opened 2023-04-28
Regression amend test

imranbashir01:
- confirm SR page
- amend journey
- order summary page
- extraction of order id


imranbashir01:
raising different pullrequest

PR #1104, opened 2023-05-03
Feature/21767 save filter query

dssherlock:


bjss-colin:
Closing this for now. The approach is a good foundation.

PR #1108, opened 2023-05-09
Additional filters by Client application type

DeepikaDK001:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/Filters/AdditionalFiltersModel.cs
```diff
@@ -11,18 +16,34 @@ public AdditionalFiltersModel()
...
         public string SelectedFrameworkId { get; set; }
 
         public List<SelectOption<string>> FrameworkOptions { get; set; }
+
+        public List<SelectOption<int>> ClientApplicaitontypeOptions { get; set; }
+
+        public void GetClientApplicationType(string clientApplicationTypeSelected)
+        {
+            ClientApplicaitontypeOptions = Enum.GetValues(typeof(ClientApplicationType))
+            .Cast<ClientApplicationType>()
```
ghost: Very nice - I like the usage of `.Cast<ClientApplicationType>` here.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Shared/_AdditionalFilters.cshtml
```diff
@@ -11,6 +14,26 @@
                                display-name="Text"/>
         </nhs-fieldset-form-label>
     </nhs-expander>
+     <nhs-expander label-text="Client application Type" open="true">
+        <nhs-table>
+            @foreach (var value in Model.ClientApplicaitontypeOptions)
```
ghost: Use a `for` loop rather than a `foreach` loop.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Shared/_AdditionalFilters.cshtml
```diff
@@ -1,5 +1,8 @@
 ﻿@model NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Solutions.Models.Filters.AdditionalFiltersModel
 <form method="post">
+    @{
+        var clientAppTypeIndex = 0;
```
ghost: This can be removed by using a `for` loop below.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Shared/_AdditionalFilters.cshtml
```diff
@@ -11,6 +14,26 @@
                                display-name="Text"/>
         </nhs-fieldset-form-label>
     </nhs-expander>
+     <nhs-expander label-text="Client application Type" open="true">
+        <nhs-table>
+            @foreach (var value in Model.ClientApplicaitontypeOptions)
+            {
+                <nhs-table-row-container>
```
ghost: Checkbox components should live inside an `nhs-fieldset-form-label` and not a table. Have a peek at some other in-code examples of `nhs-fieldset-form-label` and `nhs-checkbox-container`

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -6,9 +6,11 @@
 using System.Threading.Tasks;
 using Microsoft.Data.SqlClient;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.IdentityModel.Tokens;
```
ghost: This namespace import shouldn't be needed since it doesn't look like any code uses it.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/Filters/AdditionalFiltersModel.cs
```diff
@@ -11,18 +16,34 @@ public AdditionalFiltersModel()
...
                 Selected = false,
             }).ToList();
+            GetClientApplicationType(clientApplicationTypeSelected);
         }
 
         public string SelectedFrameworkId { get; set; }
 
         public List<SelectOption<string>> FrameworkOptions { get; set; }
+
+        public List<SelectOption<int>> ClientApplicaitontypeOptions { get; set; }
```
ghost: Spelling mistake and casing -

`ClientApplicationTypeOptions`

tests/NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests/Areas/Solutions/Models/Filters/AdditionalFilterModelTests.cs
```diff
@@ -29,5 +30,25 @@ public static void Constructor_WithFrameworks_CreatesFrameworkOptions(List<Frame
...
+        public static void Constructor_WithClientApplicationTypeSelected_CreatesClientApplicationTypeCheckBoxItems(string clientApplicationTypeSelected)
+        {
+            var model = new AdditionalFiltersModel(new List<FrameworkFilterInfo>(), clientApplicationTypeSelected);
+
+            model.ClientApplicaitontypeOptions.Should().NotBeNull();
+            model.ClientApplicaitontypeOptions.Should().HaveCount(3);
+
+            foreach (var item in model.ClientApplicaitontypeOptions)
+            {
+                if (clientApplicationTypeSelected != null && clientApplicationTypeSelected.Contains(item.Value.ToString()))
```
ghost: Remember that tests should be written in an assertive fashion. It's not immediately clear whether this test will pass with all `Selected` values being true or false. Or whether the test will fail because one `Selected` value isn't as expected.

Typically if a test contains an `if` statement then it indicates that the test should be split up into two separate tests as there are different edge cases being tested.

Be sure to identify the edge cases of the system under test, in this case `AdditionalFiltersModel`, and write one test per edge case. While also ensuring that the tests follow the [Arrange, Act, Assert pattern](https://freecontent.manning.com/making-better-unit-tests-part-1-the-aaa-pattern/)

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -59,6 +62,32 @@ public sealed class SolutionsFilterService : ISolutionsFilterService
 
             if (!string.IsNullOrWhiteSpace(selectedFrameworkId))
                 query = query.Where(ci => ci.Solution.FrameworkSolutions.Any(fs => fs.FrameworkId == selectedFrameworkId));
+            if (!string.IsNullOrWhiteSpace(clientApplicationTypeSelected))
+            {
+                string[] clientApplicationtypes = clientApplicationTypeSelected.Split(',');
```
ghost: Let's refactor this to clean up a bit. Lines 67-88 can be split out to an appropriately named method.

ghost: The Client Application Types can be parsed as enums at this point. Add some LINQ statements after the `.Split(','`) call in order to convert these from string to enum.

src/NHSD.GPIT.BuyingCatalogue.Framework/Models/SelectOption.cs
```diff
@@ -38,7 +42,10 @@ public SelectOption(string text, string advice, TValue value)
 
         public static bool operator ==(SelectOption<TValue> left, SelectOption<TValue> right)
         {
-            return left.Equals(right);
+            if (left is null)
+                return right is null;
+            else
```
ghost: `else` is redundant here since the `if` returns from the method.

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -59,6 +62,32 @@ public sealed class SolutionsFilterService : ISolutionsFilterService
...
+                foreach (var row in query)
+                {
+                    if (!string.IsNullOrEmpty(row.Solution.ClientApplication))
+                    {
+                        var clientApplication = JsonDeserializer.Deserialize<ClientApplication>(row.Solution.ClientApplication);
+                        int presentCount = 0;
```
ghost: The current `presentCount` logic means this behaves as an _**OR**_. As in, it searches for Client App Type A _OR_ B rather than A _AND_ B.

I appreciate it's not easy because the Client Application Type schema is, well, terrible.

In order to do this as an _AND_ then lines 73-82 can be replaced with a LINQ `.Where` statement on the `clientApplicationTypes` array from line #67.

The `.Where` should use `HasClientApplicationType`. You can then use `.Count` on this result. If the `.Count` is less than the number of elements in `clientApplicationTypes` then that means the solution does not reference all selected Client Application Types and the `query.Where` statement should be updated accordingly to exclude the row ID

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/SolutionsController.cs
```diff
@@ -447,5 +456,12 @@ public IActionResult AboutPilotSolutions()
 
             return View(model);
         }
+
```
dssherlock: Could this be moved into the AdditionalFiltersModel as a Get property?

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -43,12 +44,41 @@ public sealed class SolutionsFilterService : ISolutionsFilterService
             return (query, count);
         }
 
+        public IQueryable<CatalogueItem> GetClientApplicationTypeFilterQuery(IQueryable<CatalogueItem> query, string clientApplicationTypeSelected)
+        {
+            string[] clientApplicationTypes = clientApplicationTypeSelected?.Split(',');
```
ghost: These can be combined since `clientApplicationTypes` is only used in one place. Let's chain the `.Select` to the end, but also let's add a `.Where` (before the `.Select`) that uses `Enum.IsDefined` in order to only parse valid entries, since the user could manipulate the URL manually.

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -43,12 +44,41 @@ public sealed class SolutionsFilterService : ISolutionsFilterService
...
+            var clientApplicationTypeEnums = clientApplicationTypes.Select(t => (ClientApplicationType)Enum.Parse(typeof(ClientApplicationType), t));
+            if (query != null)
```
ghost: Let's move this if condition to the top of the method to reduce the level of nesting in the code. If the query is null then the rest of the method shouldn't execute.

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -43,12 +44,41 @@ public sealed class SolutionsFilterService : ISolutionsFilterService
...
+            {
+                foreach (var row in query)
+                {
+                    if (!string.IsNullOrEmpty(row.Solution.ClientApplication))
```
ghost: Similarly here, in order to reduce nesting if the `ClientApplication` is null/empty then update the query and `continue`.

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/Filters/AdditionalFiltersModel.cs
```diff
@@ -11,18 +14,44 @@ public AdditionalFiltersModel()
...
+        public void GetClientApplicationType(string clientApplicationTypeSelected)
+        {
+            ClientApplicationTypeOptions = Enum.GetValues(typeof(ClientApplicationType))
+            .Cast<ClientApplicationType>()
+            .Select(x => new SelectOption<int>
+            {
+                 Value = (int)x,
+                 Text = x.Name(),
+                 Selected = !string.IsNullOrEmpty(clientApplicationTypeSelected) && clientApplicationTypeSelected.Contains(((int)x).ToString()),
+            })
```
ghost: This list needs to be ordered alphabetically

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/SolutionsController.cs
```diff
@@ -43,7 +44,8 @@ public sealed class SolutionsController : Controller
             [FromQuery] string selectedCapabilityIds,
             [FromQuery] string selectedEpicIds,
             [FromQuery] string search,
-            string selectedFrameworkId)
+            string selectedFrameworkId,
```
dssherlock, line 47: I wonder if this parameter should be selectedClientApplicationTypeIds (to match the capability and epic naming). I know its a bit of a mouthful but currently it looks like its just a single type

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/Filters/AdditionalFiltersModel.cs
```diff
@@ -11,18 +14,45 @@ public AdditionalFiltersModel()
...
         public string SelectedFrameworkId { get; set; }
 
         public List<SelectOption<string>> FrameworkOptions { get; set; }
+
+        public List<SelectOption<int>> ClientApplicationTypeOptions { get; set; }
+
+        public string SelectedClientApplicationTypes
+        {
+            get
+            {
```
dssherlock, line 47: Should this use the FilterConstants.Delimiter as per the Epic and Capability IDs?

DeepikaDK001: yes that makes sense ..

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -43,12 +45,36 @@ public sealed class SolutionsFilterService : ISolutionsFilterService
...
+        public IQueryable<CatalogueItem> GetClientApplicationTypeFilterQuery(IQueryable<CatalogueItem> query, string selectedClientApplicationTypeIds)
+        {
+            if (query == null)
+                return query;
+
+            var clientApplicationTypeEnums = selectedClientApplicationTypeIds?.Split(FilterConstants.Delimiter).Select(t => (ClientApplicationType)Enum.Parse(typeof(ClientApplicationType), t));
+            foreach (var row in query)
+            {
+                if (string.IsNullOrEmpty(row.Solution.ClientApplication))
+                    continue;
```
ghost, line 50: Shouldn't this also also exclude this record because it doesn't have a ClientApplication?

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Solutions/ISolutionsFilterService.cs
```diff
@@ -19,7 +19,10 @@ public interface ISolutionsFilterService
             string selectedCapabilityIds = null,
             string selectedEpicIds = null,
             string search = null,
-            string selectedFrameworkId = null);
+            string selectedFrameworkId = null,
+            string selectedClientApplicationTypeIds = null);
+
+       IQueryable<CatalogueItem> GetClientApplicationTypeFilterQuery(IQueryable<CatalogueItem> query, string selectedClientApplicationTypeIds);
```
ghost: This doesn't need to be part of the interface because it isn't actually used anywhere other than in the `SolutionsFilterService`

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
DeepikaDK001: Sorry mistakenly removed it..

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -32,6 +34,32 @@ public sealed class SolutionsFilterService : ISolutionsFilterService
         public SolutionsFilterService(BuyingCatalogueDbContext dbContext) =>
             this.dbContext = dbContext ?? throw new ArgumentNullException(nameof(dbContext));
 
+        public static IQueryable<CatalogueItem> GetClientApplicationTypeFilterQuery(IQueryable<CatalogueItem> query, string selectedClientApplicationTypeIds)
+        {
+            if (query == null)
+                return query;
+
+            var clientApplicationTypeEnums = selectedClientApplicationTypeIds?.Split(FilterConstants.Delimiter).Select(t => (ClientApplicationType)Enum.Parse(typeof(ClientApplicationType), t));
```
ghost: Don't forget about the suggested `.Where` statement, after the `.Split`, which uses `Enum.IsDefined` in order to only parse valid results.

At the moment a user can manipulate the URL and type any values in which would cause this code to throw an exception. Try it out locally by changing the values in the URL.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/Filters/AdditionalFiltersModel.cs
```diff
@@ -11,18 +15,45 @@ public AdditionalFiltersModel()
...
+                return string.Join(
+                    FilterConstants.Delimiter,
+                    ClientApplicationTypeOptions.Where(x => x.Selected).Select(x => x.Value));
+            }
+        }
+
+        public void GetClientApplicationType(string selectedClientApplicationTypeIds)
+        {
+            ClientApplicationTypeOptions = Enum.GetValues(typeof(ClientApplicationType))
```
ghost: Since this code is now greatly reduced, this can be moved to the constructor and the method can be deleted

bjss-colin (changes requested):

imranbashir01 (approved):

bjss-colin (approved):

PR #1110, opened 2023-05-11
Feature/21767 save filter query

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Models/ClientApplicationType.cs
```diff
@@ -7,14 +7,14 @@ public enum ClientApplicationType
...
         [Display(Name = "Mobile or tablet")]
         [EnumMember(Value = "native-mobile")]
-        MobileTablet,
+        MobileTablet = 2,
 
         [Display(Name = "Desktop")]
         [EnumMember(Value = "native-desktop")]
-        Desktop,
+        Desktop = 3,
     }
```
DeepikaDK001: It is worth noting that for the enums ClientApplicationType and HostingType, the Enum.TryParse() method can be used to parse Integer values, even if these values are not explicitly defined within the Enum.

ghost: Exactly that. We can't change the value of the enums because it'll affect the interpretation of data. The values behind an Enum default at 0, so BrowserBased is 0 while MobileTablet is 1. After this change it's possible that the application will treat all `MobileTablet` applications as `BrowserBased`, and `Desktop` will be treated as `MobileTablet`.

Same applies to `HostingType` too.

imranbashir01 (dismissed):

database/NHSD.GPITBuyingCatalogue.Database/NHSD.GPITBuyingCatalogue.Database.sqlproj
```diff
@@ -68,6 +68,8 @@
   <Import Condition="'$(SQLDBExtensionsRefPath)' != ''" Project="$(SQLDBExtensionsRefPath)\Microsoft.Data.Tools.Schema.SqlTasks.targets" />
```
ghost, line 68: Let's add some Post-Deployment scripts to populate the database with test data too

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Filtering/Configuration/FilterEntityTypeConfiguration.cs
```diff
@@ -0,0 +1,54 @@
...
+            builder.Property(f => f.Description)
+                .IsRequired()
+                .HasMaxLength(255);
+
+            builder.Property(f => f.OrganisationId)
+                .HasConversion<int>()
+                .IsRequired();
+
+            builder.Property(f => f.FrameworkId)
+                .HasMaxLength(10);
```
ghost: This should be 36, based on the recent changes for Framework Management Admin pages.

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Filtering/Models/Filter.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using NHSD.GPIT.BuyingCatalogue.EntityFramework.Catalogue.Models;
+using NHSD.GPIT.BuyingCatalogue.EntityFramework.Organisations.Models;
+using NHSD.GPIT.BuyingCatalogue.EntityFramework.Users.Models;
+
+namespace NHSD.GPIT.BuyingCatalogue.EntityFramework.Filtering.Models
+{
+    [Serializable]
+    public partial class Filter : IAudited
```
ghost, line 10: `IAudited` could be preserved on `FIlter` in order to know who created it, but it's not necessary on `FilterCapability`, `FilterEpic`, `FilterHostingType`, `FilterClientApplicationType` since these can't change once a fitler is created.

src/NHSD.GPIT.BuyingCatalogue.Services/ServiceHelpers/SolutionsFilterHelper.cs
```diff
@@ -16,5 +17,17 @@ public static class SolutionsFilterHelper
         public static ICollection<string> ParseEpicIds(string epicIds) =>
             epicIds?.Split(FilterConstants.Delimiter, StringSplitOptions.RemoveEmptyEntries & StringSplitOptions.TrimEntries)
         .ToList() ?? new List<string>();
+
+        public static ICollection<ClientApplicationType> ParseClientApplicationTypeIds(string clientApplicationTypeIds) =>
+            clientApplicationTypeIds?.Split(FilterConstants.Delimiter, StringSplitOptions.RemoveEmptyEntries & StringSplitOptions.TrimEntries)
+                .Where(x => Enum.TryParse(typeof(ClientApplicationType), x, out var catValue) && Enum.IsDefined(typeof(ClientApplicationType), catValue))
```
ghost, line 23: `TryParse` and `Enum.IsDefined` yield the same result. `IsDefined` would be better IMO since there are less allocations taking place.

src/NHSD.GPIT.BuyingCatalogue.Services/ServiceHelpers/SolutionsFilterHelper.cs
```diff
@@ -16,5 +17,17 @@ public static class SolutionsFilterHelper
...
+                .Select(t => (ClientApplicationType)Enum.Parse(typeof(ClientApplicationType), t))
+                .ToList() ?? new List<ClientApplicationType>();
+
+        public static ICollection<HostingType> ParseHostingTypeIds(string hostingTypeIds) =>
+            hostingTypeIds?.Split(FilterConstants.Delimiter, StringSplitOptions.RemoveEmptyEntries & StringSplitOptions.TrimEntries)
+                .Where(x => Enum.TryParse(typeof(HostingType), x, out var hostingValue) && Enum.IsDefined(typeof(HostingType), hostingValue))
```
ghost, line 29: Same as above

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Solutions/IManageFiltersService.cs
```diff
@@ -0,0 +1,31 @@
...
+            string name,
+            string description,
+            int organisationId,
+            List<int> capabilityIds,
+            List<string> epicIds,
+            string frameworkId,
+            List<ClientApplicationType> clientApplicationTypes,
+            List<HostingType> hostingTypes);
+
+        Task AddFilterCapabilities(int filterId, List<int> capabilityIds);
```
ghost: The `Add` methods, except `AddFilter`, don't need to be exposed via an interface since they're only used in the `ManageFiltersService` class. As far as I know, filters can't be edited once they've been saved.

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/ManageFiltersService.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            await AddFilterCapabilities(filter.Id, capabilityIds);
+            await AddFilterEpics(filter.Id, epicIds);
+            await AddFilterClientApplicationTypes(filter.Id, clientApplicationTypes);
+            await AddFilterHostingTypes(filter.Id, hostingTypes);
+
+            return filter.Id;
+        }
+
+        public async Task AddFilterCapabilities(int filterId, List<int> capabilityIds)
```
ghost: `Add` methods, except `AddFilter`, can be marked as private here - see comment on the interface type. Alternatively mark as `internal` if they're unit tested.

database/NHSD.GPITBuyingCatalogue.Database/Filtering/Tables/FilterHostingTypes .sql
```diff
@@ -0,0 +1,14 @@
+﻿CREATE TABLE filtering.FilterHostingTypes
```
ghost: There's a space in this filename between `FilterHostingTypes` and `.sql`

bjss-colin (changes requested):

imranbashir01 (dismissed):

bjss-colin:
Note that I did hput some comments regarding temporal tables, which you may see in your emails. I've since deleted these as we can actually use the Temporal functionality of the Filters as part of Gen 2.

In other words, let's stick with the usage of Temporal tables.

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Filtering/Configuration/FilterCapabilityEntityTypeConfiguration.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using NHSD.GPIT.BuyingCatalogue.EntityFramework.Filtering.Models;
+
+namespace NHSD.GPIT.BuyingCatalogue.EntityFramework.Filtering.Configuration
+{
+    internal sealed class FilterCapabilityEntityTypeConfiguration : IEntityTypeConfiguration<FilterCapability>
+    {
+        public void Configure(EntityTypeBuilder<FilterCapability> builder)
+        {
+            builder.ToTable("FilterCapabilities", Schemas.Filtering);
```
ghost: Since these have support for Temporal tables, all `ToTable` calls will need a third parameter of `b => b.IsTemporal()`. There are some examples of this available in other Configuration classes.

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/ManageFiltersService.cs
dssherlock: They originally were all private. But then i tried to be clever and make them public for the manage filters ticket!

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Solutions/IManageFiltersService.cs
dssherlock: See above! Easy to swap back

src/NHSD.GPIT.BuyingCatalogue.Services/ServiceHelpers/SolutionsFilterHelper.cs
dssherlock: TryParse returns success (and a value) for values that haven't actually been defined. E.g. if gave the function the string 1.6 it would return 2 values rather than filtering out the 6 which isnt actually defined as anything.

src/NHSD.GPIT.BuyingCatalogue.Services/ServiceHelpers/SolutionsFilterHelper.cs
dssherlock: Same as above!

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/ManageFiltersService.cs
bjss-colin: I may be out of the loop in terms of requirements, but what will the Manage Filters ticket be doing? Capabilities, Epics and Additional Filters can't be changed by the user once their filter is created, so the writing is a one time operation that takes place via  `AddFilter`.

All of the methods can still exist, they just don't need to be public unless the requirements on adding/editing filters has changed.

src/NHSD.GPIT.BuyingCatalogue.Services/ServiceHelpers/SolutionsFilterHelper.cs
bjss-colin: `IsDefined` does have an overload which accepts an object rather than the specific `Enum` type. Hence, parsing isn't required in order to call `IsDefined`. In this case parsing occurs twice, once in `.Where` and again in `.Select`

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.Services/ServiceHelpers/SolutionsFilterHelper.cs
ghost: `IsDefined` does have an overload which accepts an object rather than the specific `Enum` type. Hence, parsing isn't required in order to call `IsDefined`. In this case parsing occurs twice, once in `.Where` and again in `.Select`

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/ManageFiltersService.cs
ghost: I may be out of the loop in terms of requirements, but what will the Manage Filters ticket be doing? Capabilities, Epics and Additional Filters can't be changed by the user once their filter is created, so the writing is a one time operation that takes place via AddFilter.

All of the methods can still exist, they just don't need to be public unless the requirements on adding/editing filters has changed.

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Filtering/Configuration/FilterCapabilityEntityTypeConfiguration.cs
dssherlock: There only seem to be 2 uses of this? The Suppliers table and the AspNetUsers table?

dssherlock: Why are these special?!

ghost: `IEntityTypeConfiguration<T>` is applied during model building in EF Core and tells EF Core how the model is made up.

More importantly, if we use a code-first approach, as we do in the Integration tests (`EnsureCreated`), and EF Core isn't aware of temporal tables then any Temporal queries will fail.

src/NHSD.GPIT.BuyingCatalogue.Services/ServiceHelpers/SolutionsFilterHelper.cs
dssherlock: But you would still need to do an int.parse? Enum.IsDefined(typeof(ClientApplicationType), "1") returns false because u are passing it "1" not 1. It would work if you passed it the string name.

ghost: Good point actually. Just tested this locally too. In that case it looks like the code in `SolutionsFilterService` is somehow behaving correctly despite not parsing first. We can revisit that one separately.

bjss-colin (approved):

samdasari9 (approved):

PR #1113, opened 2023-05-11
Fix the previous order totals for subsequent amends

martinkimmings:
This PR removes the `TotalCost` extension method for an `order` in favour off the `orderWrapper` methods `TotalCost` and  `TotalPreviousCost`. 

Calls like `OrderWrapper.Previous.TotalCost()` are replaced with `OrderWrapper.TotalPreviousCost()` which carries out the calculation for the original order (using the maximum term) and each revision (using a term derived from the planned delivery dates).

### DIFF
--- a/src/NHSD.GPIT.BuyingCatalogue.Framework/Calculations/CataloguePriceCalculations.cs
+++ b/src/NHSD.GPIT.BuyingCatalogue.Framework/Calculations/CataloguePriceCalculations.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.AspNetCore.Mvc.Formatters;
 using NHSD.GPIT.BuyingCatalogue.EntityFramework.Catalogue.Models;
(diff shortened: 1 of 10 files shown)


src/NHSD.GPIT.BuyingCatalogue.Framework/Calculations/CataloguePriceCalculations.cs
```diff
@@ -101,12 +100,14 @@ public static decimal TotalAnnualCost(this Order order, bool roundResult = false
             return total;
         }
 
-        public static decimal TotalCost(this Order order, bool roundResult = false)
+        public static decimal TotalPreviousCost(this OrderWrapper orderWrapper, bool roundResult = false)
         {
-            var maximumTerm = order?.MaximumTerm ?? 36;
+            if (orderWrapper == null)
+            {
+                return decimal.Zero;
```
ghost, line 107: For the sake of completeness, let's get a unit test added ot test this edge case.

bjss-colin (changes requested):

bjss-colin (approved):

imranbashir01 (approved):

PR #1120, opened 2023-05-17
Add buyer dashboard

bjss-colin:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Controllers/BuyerDashboardController.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    [HttpGet]
+    public async Task<IActionResult> Index()
+    {
+        var organisationId = User.GetPrimaryOrganisationInternalIdentifier();
+
+        var organisation = await organisationsService.GetOrganisationByInternalIdentifier(organisationId);
+
+        var model = new BuyerDashboardModel(
+            organisation.Name,
```
dssherlock, line 32: Won't this always be false if you have have only got the [Authorise("Buyer")] attribute on the class?

dssherlock (approved):

samdasari9 (approved):

PR #1125, opened 2023-05-18
Feature/21806 manage filters summary

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/ManageFiltersController.cs
```diff
@@ -138,6 +139,17 @@ public async Task<IActionResult> Index()
                 typeof(ManageFiltersController).ControllerName());
         }
 
+        [HttpGet("filter-details")]
+        public IActionResult FilterDetails(int filterId)
+        {
+            var model = new NavBaseModel()
+            {
+                BackLink = Url.Action(nameof(Index), typeof(ManageFiltersController).ControllerName()),
+                BackLinkText = "Go back",
```
ghost: `Go back` is the default value so no assignment to `BackLinkText` is needed here

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/ManageFiltersController.cs
```diff
@@ -77,7 +78,7 @@ public async Task<IActionResult> Index()
             string selectedClientApplicationTypeIds,
             string selectedHostingTypeIds)
         {
-            var organisationId = await GetUserOrganisationId();
+            var organisationId = (await GetUserOrganisation()).Id;
             var existingFilters = await manageFiltersService.GetFilters(organisationId);
```
ghost: `GetFilters` returns _everything_ that lives under a filter. We've previously had issues with performance when doing this, particularly when it's a dashboard

It's worth adding a new method to the `FilterService` that returns just the filter with no navigational properties because all you're really after here is the Name, Description and Last Updated date.

As a separate point, it's worth adding `AsSplitQuery` and `AsNoTracking` to the implementation of `GetFilters` in order to prevent a [cartesian explosion](https://www.thinktecture.com/en/entity-framework-core/cartesian-explosion-problem-in-3-1/).

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/ManageFilters/Index.cshtml
```diff
@@ -1,70 +1,98 @@
...
-                        </p>
-                    }
-                </nhs-summary-list-row>
-            </nhs-summary-list>
+            <p data-test-id="no-filters">No filters have been added for your organisation yet.</p>
+        }
+        else
+        {
+            var filters = Model.Filters.Count() == 1 ? "filter" : "filters";
+            <p>@Model.Filters.Count() saved search @filters.</p>
```
ghost: Since `Filters` is a `List<T>`, you can use `Count` here rather than LINQ's `Count()`

bjss-colin (changes requested):
Just a few minor points

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/ManageFiltersController.cs
dssherlock: Thinking about it, do we really need to add another method (i.e GetFiltersThin)? I think we can just replace GetFilters altogether as we only need all the details for a specific Filter on the Details page?

imranbashir01 (dismissed):

imranbashir01 (approved):

bjss-colin (approved):

PR #1147, opened 2023-06-06
Display the legal supplier name in the contract, pdf and csv

DeepikaDK001:


src/NHSD.GPIT.BuyingCatalogue.Services/Csv/CsvService.cs
```diff
@@ -175,7 +177,7 @@ private async Task<(int SupplierId, string SupplierName)> GetSupplierDetails(int
 
             output ??= order.Supplier;
 
-            return (output?.Id ?? 0, output?.Name ?? string.Empty);
+            return (output?.Id ?? 0, string.Equals(output.Name, output.LegalName, StringComparison.OrdinalIgnoreCase) ? output.Name : output.LegalName ?? string.Empty);
```
ghost: `?? string.Empty` is only being applied to the legal name. An exception will also be thrown if `output` is null since the null-conditional operator has been removed (see the differences between the previous and current files).

Maybe consider assigning `output.Name` and `output.LegalName` to two separate variables, ensuring to use the original format of the null conditional and null coalescing operators in both variables. Then use the ternary to check for equality picking the variable based on the outcome.

bjss-colin (changes requested):

samdasari9 (approved):

samdasari9 (approved):

bjss-colin (approved):

PR #1155, opened 2023-06-14
Name change from 'Client application type' to 'Application type'

DeepikaDK001:


samdasari9 (dismissed):

bjss-colin (changes requested):
There are still a few dozen references to `Client Application` in HTML and C# code.

Use CTRL+Shift+F to look for `Client Application` _**and**_ `ClientApplication`

The new Solutions Filter Results PDF is one example

PR #1167, opened 2023-06-23
Regression tests

imranbashir01:
-All of the service recipients were being selected for the multiple service selection scenarios which was impacting the run time of the full suite, changed it to select only handful of the service recipients and now the run time has improved a lot.
-put in another test to cover the select all service recipients scenario


tests/NHSD.GPIT.BuyingCatalogue.E2ETests.Framework/Actions/Common/CommonActions.cs
```diff
@@ -53,6 +54,9 @@ public void ClickLinkElement(By targetElement, string hrefContains)
         public void ClickAllCheckboxes() =>
             Driver.FindElements(By.CssSelector("input[type=checkbox]")).ToList().ForEach(element => element.Click());
 
+        public void ClickMultipleCheckboxes() =>
+            Driver.FindElements(By.CssSelector("input[type=checkbox]")).ToList().Take(3).ForEach(element => element.Click());
```
ghost: Pass this as a parameter instead. That way you can select one or may checkboxes in different tests

tests/NHSD.GPIT.BuyingCatalogue.E2ETests.Framework/Actions/Common/CommonActions.cs
```diff
@@ -262,6 +266,11 @@ public bool CheckBoxSelectedByLabelText(string labelText)
             .FindElements(By.ClassName("nhsuk-checkboxes__input"))
             .All(cb => cb.Selected);
 
+        public bool MultipleCheckboxeSelected() =>
```
ghost: Spelling mistake `MultipleCheckboxeSelected`. It might be worth renaming this as it checks to see if there are _any_ checkboxes selected, rather than there being multiple selected.

bjss-colin (changes requested):

bjss-colin (approved):

PR #1174, opened 2023-07-06
Gen2 import and CompliancyLevelId schema change

martinkimmings:
This PR adds navigations to the new `CapabiltyEpics` join entity type from `Capability` and `Epic` in addition to the existing skip navigations for the many to many relationship between `Capability` and `Epic`

> [In a sense, this mapping provides the best of both worlds between a simple many-to-many mapping, and a mapping that more explicitly matches the database schema.](https://learn.microsoft.com/en-us/ef/core/modeling/relationships/many-to-many#many-to-many-with-navigations-to-join-entity)

We need this additional control over the join entity so that we can set and query `CompliancyLevel`

> [It is also common to create navigations to the join entity in addition to the "skip navigations" used for the many-to-many relationship. These additional navigations allow the join entity to be easily referenced from code, thereby facilitating reading and/or changing the payload data](https://learn.microsoft.com/en-us/ef/core/modeling/relationships/many-to-many#many-to-many-and-join-table-with-payload)

`CompliancyLevelId` moved to `CompatabilityEpics`

image


database/NHSD.GPITBuyingCatalogue.Database/Catalogue/Tables/CapabilityEpics.sql
```diff
@@ -2,10 +2,12 @@
 (
     [CapabilityId] INT NOT NULL,
     [EpicId] NVARCHAR(10) NOT NULL,
+    CompliancyLevelId int NULL,
```
ghost, line 5: Let's add an idempotent Post-Deployment script to handle the data migration for migrating all existing epics to this new CompliancyLevel format. The post deployment script should only run once based on whether there are any rows with this column being empty.

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Models/Capability.Custom.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace NHSD.GPIT.BuyingCatalogue.EntityFramework.Catalogue.Models
+{
+    public sealed partial class Capability
```
ghost: I've personally never been a fan of this `.Custom.cs` convention that has been used in this codebase, since it doesn't yield much of a benefit, and have been removing any I come across. These methods could live in the `Capability` class directly.

bjss-colin (changes requested):
Two minor points

database/NHSD.GPITBuyingCatalogue.Database/Catalogue/Tables/CapabilityEpics.sql
martinkimmings: I've added `database/NHSD.GPITBuyingCatalogue.Database/PostDeployment/MigrateEpicCompliancyLevel.sql`

database/NHSD.GPITBuyingCatalogue.Database/PostDeployment/MigrateEpicCompliancyLevel.sql
```diff
@@ -0,0 +1,7 @@
+IF NOT EXISTS(SELECT 1 FROM catalogue.CapabilityEpics WHERE CompliancyLevelId IS NOT NULL)
+  UPDATE ce
+    SET ce.CompliancyLevelId = e.CompliancyLevelId  
+	FROM catalogue.CapabilityEpics ce
+      JOIN catalogue.epics e
+	    ON e.CapabilityId = ce.CapabilityId AND e.id = ce.EpicId
```
ghost: Just the Epic ID should be fine for now. The problem is that the CapabilityId column on the Epics table hasn't been in use since the 2nd or 3rd last release. There may be null values in the production database for this.

Since Epics are still a one to one, wrapped as a one to many, I don't think using just the Epic Id will cause any problems.

bjss-colin (changes requested):
One more point, otherwise everything else looks good.

bjss-colin (approved):

imranbashir01 (approved):

PR #1175, opened 2023-07-07
Feature/2a associated service bespoke milestones

dssherlock:
image


database/NHSD.GPITBuyingCatalogue.Database/Ordering/Tables/ContractBilling.sql
```diff
@@ -0,0 +1,13 @@
+﻿CREATE TABLE ordering.ContractBilling
```
ghost, line 1: This is a link table so I'm not sure if we need temporal functionality. What would be the use case?

database/NHSD.GPITBuyingCatalogue.Database/Ordering/Tables/ContractBillingItems.sql
```diff
@@ -0,0 +1,17 @@
+﻿CREATE TABLE ordering.ContractBillingItems
```
ghost, line 1: Also worth considering whether this needs temporalness. We should always evaluate the use case on a table by table basis when adding new ones.

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractBillingService.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        public ContractBillingService(BuyingCatalogueDbContext dbContext)
+        {
+            this.dbContext = dbContext ?? throw new ArgumentNullException(nameof(dbContext));
+        }
+
+        public async Task<Contract> AddContractBilling(int orderId, int contractId)
+        {
+            var contract = await dbContext.Contracts
+                .Include(x => x.Order)
```
ghost: These includes are only required in `AddBespokeContractBillingItem`, whereas `ContractBillingController` also calls this method and discards the Order, OrderItems, CatalogueItem and ContractBilling.

Let's use a separate query in `AddBespokeContractBillingItem` and make use of the `??=` operator to conditionally assign a new instance of `ContractBilling` if it is indeed null.

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractBillingService.cs
```diff
@@ -0,0 +1,128 @@
...
+                    .ThenInclude(x => x.OrderItems)
+                        .ThenInclude(x => x.CatalogueItem)
+                .Include(x => x.ContractBilling)
+                .FirstOrDefaultAsync(o => o.Id == contractId && o.OrderId == orderId) ?? throw new ArgumentException("Invalid contract", nameof(contractId));
```
ghost: Is there a scneario where this Contract could ever be null if the user is this far? Let's remove this exception and handle accordingly.

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractBillingService.cs
```diff
@@ -0,0 +1,128 @@
...
+        public async Task AddBespokeContractBillingItem(int orderId, int contractId, CatalogueItemId catalogueItemId, string name, string paymentTrigger, int quantity)
+        {
+            if (string.IsNullOrEmpty(name))
+                throw new ArgumentNullException(nameof(name));
+            if (string.IsNullOrEmpty(paymentTrigger))
+                throw new ArgumentNullException(nameof(paymentTrigger));
+
+            var contract = await AddContractBilling(orderId, contractId);
+
+            var associatedService = contract.Order.GetAssociatedService(catalogueItemId) ?? throw new ArgumentException("Invalid associated service", nameof(catalogueItemId));
```
ghost: Same as above regarding the exception

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractBillingService.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            if (item == null)
+            {
+                return;
+            }
+
+            var order = await dbContext.Orders
+                    .Include(x => x.OrderItems)
+                    .ThenInclude(x => x.CatalogueItem)
+                .FirstOrDefaultAsync(x => x.Id == orderId) ?? throw new ArgumentException("Invalid order", nameof(orderId));
```
ghost: Same as above regarding exception

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractBillingService.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            var associatedService = order.GetAssociatedService(catalogueItemId) ?? throw new ArgumentException("Invalid associated service", nameof(catalogueItemId));
```
ghost: Same as above regarding exception

ghost: You could go directly to the dbContext.OrderItems table to retrieve the associated service by Order and Item ID, since we already have this information at hand via the method parameters

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractBillingService.cs
```diff
@@ -0,0 +1,128 @@
...
+            dbContext.ContractBillingItems.Remove(item);
+            await dbContext.SaveChangesAsync();
+        }
+
+        public async Task DeleteContractBillingItems(int orderId, IEnumerable<CatalogueItemId> catalogueItemIds)
+        {
+            var items = dbContext.ContractBillingItems
+                .Where(x => x.OrderId == orderId && catalogueItemIds.Contains(x.CatalogueItemId));
+
+            if (items.Any())
```
ghost: Let's invert this if statement to reduce nesting

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractBillingService.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        public async Task AddBespokeContractBillingItem(int orderId, int contractId, CatalogueItemId catalogueItemId, string name, string paymentTrigger, int quantity)
+        {
+            if (string.IsNullOrEmpty(name))
+                throw new ArgumentNullException(nameof(name));
+            if (string.IsNullOrEmpty(paymentTrigger))
+                throw new ArgumentNullException(nameof(paymentTrigger));
+
+            var contract = await AddContractBilling(orderId, contractId);
```
ghost: Is it possible that the `ContractBilling` object won't exist at this time? The Controller flow looks like it will exist.

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractsService.cs
```diff
@@ -19,34 +20,60 @@ public ContractsService(BuyingCatalogueDbContext dbContext)
 
         public async Task<Contract> GetContract(int orderId)
         {
-            return await dbContext.Contracts
+            var contract = await dbContext.Contracts
+                .AsNoTracking()
+                .FirstOrDefaultAsync(x => x.OrderId == orderId);
+
+            var output = await AddContract(contract, orderId);
```
ghost, line 27: I understand the intent behind this, but it's a bit confusing to see calls to `AddContract` in the `GetContract` method. since Get methods should usually be idempotent whereas this will have side-effects. This should return null if there's no contract, or a valid contract if there is one.

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractsService.cs
```diff
@@ -19,34 +20,60 @@ public ContractsService(BuyingCatalogueDbContext dbContext)
...
+        public async Task<Contract> GetContractWithImplementationPlan(int orderId)
+        {
+            var contract = await dbContext.Contracts
                 .AsNoTracking()
                 .Include(x => x.ImplementationPlan)
-                .ThenInclude(x => x.Milestones.OrderBy(m => m.Order))
+                    .ThenInclude(x => x.Milestones.OrderBy(m => m.Order))
                 .FirstOrDefaultAsync(x => x.OrderId == orderId);
+
+            var output = await AddContract(contract, orderId);
```
ghost, line 40: Same as above

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractsService.cs
```diff
@@ -19,34 +20,60 @@ public ContractsService(BuyingCatalogueDbContext dbContext)
...
+                    .ThenInclude(x => x.ContractBillingItems)
+                        .ThenInclude(x => x.OrderItem)
+                            .ThenInclude(x => x.CatalogueItem)
                 .FirstOrDefaultAsync(x => x.OrderId == orderId);
 
-            if (output != null)
-            {
-                return output;
-            }
+            var output = await AddContract(contract, orderId);
```
ghost, line 59: Same as above

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ImplementationPlanService.cs
```diff
@@ -30,23 +27,30 @@ public async Task<ImplementationPlan> GetDefaultImplementationPlan()
...
+        public async Task<Contract> AddImplementationPlan(int orderId, int contractId)
         {
-            if (string.IsNullOrEmpty(name))
-                throw new ArgumentNullException(nameof(name));
-            if (string.IsNullOrEmpty(paymentTrigger))
-                throw new ArgumentNullException(nameof(paymentTrigger));
-
             var contract = await dbContext.Contracts.Include(x => x.ImplementationPlan)
-                    .FirstOrDefaultAsync(o => o.Id == contractId && o.OrderId == orderId) ?? throw new ArgumentException("Invalid contract", nameof(contractId));
+                .FirstOrDefaultAsync(o => o.Id == contractId && o.OrderId == orderId) ?? throw new ArgumentException("Invalid contract", nameof(contractId));
```
ghost: No need for an exception here if the contract isn't null, or can't be null at this point.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/ContractBilling/DeleteContractBillingItem.cshtml
```diff
@@ -0,0 +1,34 @@
...
+    <div class="nhsuk-grid-column-full">
+        
+        <nhs-page-title title="@ViewBag.Title"
+                        caption="@Model.ItemName"
+                        advice="The bespoke milestone will be deleted from this order." />
+
+        <form method="post">
+            <div class="nhsuk-u-margin-top-9">
+
+                <input type="hidden" asp-for="CallOffId" />
```
ghost, line 21: We'll need to add the backlink as a hidden input also also

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/NHSD.GPIT.BuyingCatalogue.IntegrationTests.csproj
```diff
@@ -60,4 +60,8 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup>
```
ghost: Can be removed

bjss-colin (changes requested):

database/NHSD.GPITBuyingCatalogue.Database/Ordering/Tables/ContractBilling.sql
dssherlock: The existence of a ContractBilling on the contract indicates that the Default milestones have been accepted. Felt this was worth keeping track of. As you say the default at present is to add them so followed the convention. Happy to remove if you want to start getting rid of them

database/NHSD.GPITBuyingCatalogue.Database/Ordering/Tables/ContractBillingItems.sql
dssherlock: Yep similar to above. The existing table ImplementationPlanMilestones has the temporal tables associated. This is an equivalent table so felt if it was deemed necessary for the one then it will be for this one

database/NHSD.GPITBuyingCatalogue.Database/Ordering/Tables/ContractBilling.sql
ghost: > As you say the default at present is to add them so followed the convention

Not sure if this was something agreed/decided by the previous team, I can't recall any conversations around temporal support other than it being a thing. The default _should_ be to evaluate on a case by case basis to determine whether an entity could benefit from temporal functionality, since temporal restricts some actions and introduces further complexities with data migration. Particularly with link tables, I can't imagine many scenarios where we'd take advantage of this in terms of auditing or presentation.

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/ContractsService.cs
dssherlock: So make the AddContract public and call directly from the controller after each Get call? Feels a bit repetitive

bjss-colin (approved):

samdasari9 (approved):

PR #1181, opened 2023-07-13
Feature/us 22321 marketing item5 additional services

DeepikaDK001:


samdasari9 (approved):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/AdditionalServices.cshtml
```diff
@@ -10,90 +10,25 @@
...
-                    typeof(SolutionsController).ControllerName(),
-                    new { Model.SolutionId, additionalServiceId = service.Id }),
-                Prices = service.CataloguePrices,
-                ItemDescription = service.AdditionalService.FullDescription,
-            };
-            <partial name="_MarketingPriceSummary" model="priceSummaryModel" />
-
-            <partial name="_BackToTop" />
+            <nhs-expander label-text="@service.Name" open="false">
+                @{
```
dssherlock, line 21: It might be nicer to create this in the model instead? Not good to put lots of logic in the view if you can help it.

dssherlock (approved):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/AdditionalServices.cshtml
DeepikaDK001: I think I could see that the logic to create the priceSummaryModel is not directly present in the view. Instead it is within a code block in the view that is iterating over the Model.Services. In this case, the logic is already in for the particular service to display them separately within the expander , as it is generating the necessary data for each service being displayed in the view. Moving this specific logic to the model wouldn't provide any significant benefits since it is closely related to the view's rendering of the Services. It will be hard to use url.Action in the model as well.

PR #1194, opened 2023-07-27
Feature/bespoke milestone data migration

dssherlock:
### Milestones
image

### Requirements
image


tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Utils/Extensions/BuyingCatalogueDbContextExtensions.cs
```diff
@@ -30,5 +30,30 @@ public static ContractFlags GetContractFlags(this BuyingCatalogueDbContext conte
 
             return output;
         }
+
+        public static Contract GetContract(this BuyingCatalogueDbContext context, int orderId)
```
ghost: Let's make this asynchronous by using `FirstOrDefaultAsync` on the context. It might also be worth making this an extension method for `DbSet<Contract>` since then you can use `await context.Contracts.GetContract()` which just makes it a bit more isolated to the contracts tables.

bjss-colin (changes requested):

tests/NHSD.GPIT.BuyingCatalogue.E2E.PublicBrowseTests/Utils/Extensions/BuyingCatalogueDbContextExtensions.cs
dssherlock: Yep can do-it was just a direct copy of the existing GetContractFlags method above that is not really required anymore (or has limited usage) as I've removed code references to the flags. I didnt spend any time on it as I didnt think we were really supporting integration tests anymore.

dssherlock: Would u do the context.SaveChanges() after each call then?

imranbashir01 (approved):

bjss-colin (approved):

PR #1203, opened 2023-08-11
building blocks for single and no solution results

imranbashir01:


tests/NHSD.GPIT.BuyingCatalogue.RegressionTests/Pages/Competitions/CompetitionPages.cs
```diff
@@ -43,23 +46,38 @@ public CompetitionPages(IWebDriver driver, CommonActions commonActions, LocalWeb
...
 
             ReviewFilter.ReviewYourFilter();
 
             StartCompetition.CreateCompetition(competitionName);
 
-            int competitionId = CompetitionId();
-            int filterId = GetFilterId(competitionId);
-
-            if (filterId == 2)
+            if (filterType == "No results filter")
```
ghost: Let's use an enum instead for the `filterType`.

You could have something like

```csharp
public enum FilterType
{
    NoResults,
    SingleResult,
    MultipleResults,
}
```

And could then select one via `FilterType.NoResult`, `FilterType.SingleResult` or `FilterType.MultipleResults`.

This constrains consumers of this method to pass a valid and supported input and is generally cleaner than relying on string comparison.

tests/NHSD.GPIT.BuyingCatalogue.RegressionTests/Pages/Competitions/CompetitionPages.cs
```diff
@@ -43,23 +46,38 @@ public CompetitionPages(IWebDriver driver, CommonActions commonActions, LocalWeb
...
         internal CompetitionStepOne CompetitionStepOne { get; }
 
+        internal NoSolutionsFound NoSolutionsFound { get; }
+
+        internal SingleSolutionFound SingleSolutionFound { get; }
+
         internal IWebDriver Driver { get; }
 
-        public void StepOnePrepareCompetition(string filterType, string competitionName, int addNumberOfSolutions)
+        public void StepOnePrepareCompetition(string filterType, string competitionName,int addNumberOfSolutions = 0)
```
ghost: With the change below, the `filterType` parameter could change from `string filterType` to `FilterType filterType`, where the capitalized `FilterType` is the enum type and lowercase `filterType` is the parameter name.

bjss-colin (changes requested):
A minor point regarding the `filterType` inputs.

bjss-colin (approved):

PR #1208, opened 2023-08-18
Additional Filters by Interoperability

DeepikaDK001:


src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Extensions/EnumExtensions.cs
```diff
@@ -13,5 +13,11 @@ public static string Name<TEnum>(this TEnum value)
 
         public static string EnumMemberName<TEnum>(this TEnum value)
             where TEnum : struct, Enum => value.AsString(EnumFormat.EnumMemberValue);
+
+        public static string GetEnumMemberValue<TEnum>(this TEnum value)
+            where TEnum : struct, Enum => value.EnumMemberName();
```
ghost: This calls `EnumMemberName` which exists directly above. This method could be removed and code could instead call `EnumMemberName` directly.

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Extensions/EnumExtensions.cs
```diff
@@ -13,5 +13,11 @@ public static string Name<TEnum>(this TEnum value)
...
+
+        public static string GtEnumMemberIntValue<TEnun>(this TEnun value)
```
ghost: Rather than an additional extension method, casting via `(int)` could be used directly as this is already being done on lines 94 and 116 of SolutionsFilterService

bjss-colin (changes requested):

samdasari9 (approved):

bjss-colin (approved):

PR #1217, opened 2023-08-25
Summary page

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.Services/Pdf/CachedOrderPdfService.cs
```diff
@@ -28,10 +28,10 @@ public async Task<MemoryStream> CreateOrderSummaryPdf(Order order)
     {
         if (order == null) throw new ArgumentNullException(nameof(order));
 
-        if (order.OrderStatus != OrderStatus.Completed)
+        if (order.OrderStatus != OrderStatus.Completed && order.OrderStatus != OrderStatus.Terminated)
             return await orderPdfService.CreateOrderSummaryPdf(order);
 
-        var blobDocument = $"{order.CallOffId.ToString()}.pdf";
+        var blobDocument = order.OrderStatus == OrderStatus.Completed ? $"{order.CallOffId.ToString()}.pdf" : $"{order.CallOffId.ToString()}-terminated.pdf";
```
ghost, line 34: Might be worth adding a switch statement here as the assumption here is that an order is terminated if it's not completed, which isn't always the case since admins can download PDFs of a deleted order from the admin side.

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.Services/Pdf/CachedOrderPdfService.cs
dssherlock: But if the order has been deleted surely it wont reach this point in code as it would have returned the CreateOrderSummaryPdf?

dssherlock: I could swap it round so that the check for Terminated is first?

samdasari9 (approved):

bjss-colin (approved):

PR #1224, opened 2023-09-01
Feature/Terminate Contract

dssherlock:


samdasari9 (dismissed):

database/NHSD.GPITBuyingCatalogue.Database/Ordering/Tables/OrderTermination.sql
```diff
@@ -0,0 +1,10 @@
+﻿CREATE TABLE ordering.OrderTermination
+(
+    Id int IDENTITY(1, 1) NOT NULL,
+    OrderId INT NOT NULL,
+    Reason NVARCHAR(1000) NOT NULL,
+    DateOfTermination DATETIME2(0) NOT NULL,
+    CONSTRAINT PK_OrderTermination PRIMARY KEY (Id),
+    CONSTRAINT FK_OrderTermination_Order FOREIGN KEY (OrderId) REFERENCES ordering.Orders(Id),
```
ghost: `ON DELETE CASCADE` might be useful here as I'm imagining a scenario similar to the recent contract deletion bug for an amended order. The buyer wouldn't be able to delete the amendment but an admin may be able to via the admin > Manage Orders section.

src/NHSD.GPIT.BuyingCatalogue.Services/Csv/CsvService.cs
```diff
@@ -26,52 +27,25 @@ public CsvService(BuyingCatalogueDbContext dbContext, IFundingTypeService fundin
...
+                    .AsNoTracking()
+                    .Where(x => x.OrderNumber == order.OrderNumber && x.Revision < order.Revision)
+                    .OrderByDescending(x => x.Revision)
+                    .Select(y => y.Id).ToListAsync();
 
-            for (int i = 0; i < items.Count; i++)
-                items[i].ServiceRecipientItemId = $"{items[i].CallOffId}-{items[i].ServiceRecipientId}-{i}";
+                foreach (var id in revisions)
+                {
+                    items.AddRange(await CreateFullOrderCsv(id));
```
ghost, line 46: Rather than looping, `CreateFullOrderCsv` could instead accept `params int[] orderIds` and do a `.Contains` check on the `OrderItemRecipents` so it does a bulk query.

bjss-colin (changes requested):

bjss-colin (approved):

PR #1235, opened 2023-09-21
Consolidate the non practice based information sent to finance

DeepikaDK001:


src/NHSD.GPIT.BuyingCatalogue.Services/Csv/CsvService.cs
```diff
@@ -71,8 +70,10 @@ public async Task<int> CreatePatientNumberCsvAsync(int orderId, MemoryStream str
                         OdsCode = or.Order.OrderingParty.ExternalIdentifier,
                         OrganisationName = or.Order.OrderingParty.Name,
                         CommencementDate = or.Order.CommencementDate,
-                        ServiceRecipientId = or.OdsCode,
-                        ServiceRecipientName = or.OdsOrganisation.Name,
+                        ServiceRecipientId = oir.OrderItem.OrderItemPrice.CataloguePriceQuantityCalculationType != CataloguePriceQuantityCalculationType.PerServiceRecipient && (oir.OrderItem.OrderItemPrice.ProvisioningType == ProvisioningType.Declarative || oir.OrderItem.OrderItemPrice.ProvisioningType == ProvisioningType.OnDemand)
```
ghost: `IPrice`, of which `OrderItemPrice` implements, has an `IsPerServiceRecipient` method that contains similar logic as the condition block in this ternary. You'd have to cast `oir.OrderItem.OrderItemPrice` to `IPrice` but it should help reduce code by re-using existing business logic

bjss-colin (changes requested):

martinkimmings (commented):
I think it is worth adding some unit tests for the consolidation changes to show/test how multiple recipients sometimes result in a single row.

`CsvService` was recently changed to extract the dependency on temporal table data to `ISupplierTemporalService` which means testing this should be possible now.

imranbashir01 (dismissed):

src/NHSD.GPIT.BuyingCatalogue.Services/Csv/CsvService.cs
```diff
@@ -71,8 +71,10 @@ public async Task<int> CreatePatientNumberCsvAsync(int orderId, MemoryStream str
                         OdsCode = or.Order.OrderingParty.ExternalIdentifier,
                         OrganisationName = or.Order.OrderingParty.Name,
                         CommencementDate = or.Order.CommencementDate,
-                        ServiceRecipientId = or.OdsCode,
-                        ServiceRecipientName = or.OdsOrganisation.Name,
+                        ServiceRecipientId = (oir.OrderItem.OrderItemPrice as IPrice).IsPerServiceRecipient() == false && (oir.OrderItem.OrderItemPrice.ProvisioningType == ProvisioningType.Declarative || oir.OrderItem.OrderItemPrice.ProvisioningType == ProvisioningType.OnDemand)
```
ghost: How about `!(oir.OrderItem.OrderItemPrice as IPrice).IsPerServiceRecipient()`?

Don't forget to make these same changes further down in this class also, since `CreateFullOrderCsv` uses similar logic

bjss-colin (changes requested):

imranbashir01 (dismissed):

martinkimmings (approved):

imranbashir01 (approved):

bjss-colin (approved):

PR #1270, opened 2023-10-27
Feature/22563 Align PDD amend journey with normal order journey

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.Framework/Calculations/CataloguePriceCalculations.cs
```diff
@@ -144,7 +145,7 @@ private static decimal TotalCostByPlannedDelivery(this Order order, Order previo
         {
             return order.TotalOneOffCost(previous) + order?.OrderItems.Sum(i =>
             {
-                var term = GetTerm(order.EndDate, order.DetermineOrderRecipients(previous, i.CatalogueItemId), i);
+                var term = order.GetTerm();
                 return ((IPrice)i.OrderItemPrice).CalculateCostPerMonth(i.TotalQuantity(order.DetermineOrderRecipients(previous, i.CatalogueItemId))) * term;
             }) ?? decimal.Zero;
         }
```
martinkimmings: I think we can simplify this method even further now as we are using the same term for all order times similar to `TotalCostForMaximumTerm`.

dssherlock: Like this?!

imranbashir01 (dismissed):

imranbashir01 (dismissed):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Models/Contracts/DeliveryDates/OrderItemRecipientModel.cs
```diff
@@ -9,12 +10,12 @@ public OrderItemRecipientModel()
         {
         }
 
-        public OrderItemRecipientModel(OrderItemRecipient recipient)
+        public OrderItemRecipientModel(OrderRecipient recipient, CatalogueItemId catalogueItemId)
         {
-            CatalogueItemId = recipient.CatalogueItemId;
-            DeliveryDate = recipient.DeliveryDate;
+            CatalogueItemId = catalogueItemId;
+            DeliveryDate = recipient.OrderItemRecipients.FirstOrDefault(x => x.CatalogueItemId == catalogueItemId)?.DeliveryDate;
```
martinkimmings: Looks like you could call `GetDeliveryDateForItem(...)` instead.
```suggestion
            DeliveryDate = recipient.GetDeliveryDateForItem(catalogueItemId);
```

src/NHSD.GPIT.BuyingCatalogue.Framework/Calculations/CataloguePriceCalculations.cs
```diff
@@ -156,20 +153,11 @@ private static decimal TotalCostForMaximumTerm(this Order order, Order previous)
...
-            var deliveryDate = orderRecipients.Select(x => x.GetDeliveryDateForItem(orderItem.CatalogueItemId))
-                .FirstOrDefault(x => x != null);
-
-            if (!deliveryDate.HasValue) return 0;
-
-            var term = endDate.RemainingTerm(deliveryDate.Value);
-            return term;
+            return order.DeliveryDate.HasValue
+                ? order.EndDate.RemainingTerm(order.DeliveryDate.Value)
+                : order.MaximumTerm ?? 36;
```
martinkimmings: `Order.GetTerm()` is currently only called when when `Order.IsAmendment` is `true`. If this was changed to be `IsAmendment` "aware" returning the appropriate term to use I think this would allow further simplifications.

```
            if (order.IsAmendment)
            {
                return CalculateForTerm(orderItem, order.GetTerm(), recipients);
            }
            else
            {
                var maximumTerm = order.MaximumTerm ?? 36;
                return CalculateForTerm(orderItem, maximumTerm, recipients);
            }
```
Could be replaced with a single call to `CalculateForTerm(orderItem, order.GetTerm(), recipients);`

We wouldn't need the conditional logic in total `TotalCost` either as both `TotalCostByPlannedDelivery` and `TotalCostForMaximumTerm` could be removed and replaced with

```
private static decimal TotalCost(this Order order, Order previous)
{
  return order.TotalOneOffCost(previous) + (order.TotalMonthlyCost(previous) * order.GetTerm());
}
```

martinkimmings (dismissed):
Approved. I've added a few optional suggestions to consider. Happy for you to resolve as you see fit.

imranbashir01 (approved):

martinkimmings (approved):

PR #1271, opened 2023-10-30
Regression tests for further competition price only journey

imranbashir01:


tests/NHSD.GPIT.BuyingCatalogue.RegressionTests/Pages/Competitions/CompetitionPages.cs
```diff
@@ -95,16 +110,81 @@ public void StepOnePrepareCompetition(FilterType filterType, string competitionN
...
+        }
+
         private int CompetitionId()
         {
-            string url = Driver.Url;
-            int charPos = url.LastIndexOf("competitions/") + "competitions".Length + 1;
-            int charLength = url.IndexOf("/select-solutions") - charPos;
-            int competitionId = int.Parse(url.Substring(charPos, charLength));
-            return competitionId;
+            var competitionurl = Driver.Url.Split("/");
```
ghost: I think the original method was a bit safer as this could have an array out of bounds exception since you're using `i + 1`.

One option is to use string slicing, like you originally used, but instead of using `/select-solutions` you could fetch the number by going between `/competitions/` and the next instance of `/` immediately afterwards. Everything between the first and second forward-slash will be the ID.

I've commented the following code just so it's a bit easier to see what's going on

```csharp
const string lookupString = "competitions/";

// Example: https://localhost:5001/organisations/ABC/competitions/123/select-solutions"
// Spans are more efficient than using Substring. That's not too important here, I just wanted a reason to play around with Spans
var url = Driver.Url.AsSpan();

// Find the index of the beginning of the competition ID
var beginningIndex = url.IndexOf(lookupString, StringComparison.Ordinal) + lookupString.Length;

// Take a slice from the above string, rather than a substring. Example: "123/select-solutions"
var trimmedUrl = url.Slice(beginningIndex);

// Find the next forward-slash in 123/select-solutions
var endIndex = trimmedUrl.IndexOf('/');

// If the endIndex is -1 then that means the user is on the task list.
// I.E: URL is https://localhost:5001/organisations/ABC/competitions/123
// 
// If the endIndex is not -1 then we use that for the length.
var competitionIdLength = endIndex != -1 ? endIndex : trimmedUrl.Length;

// Take another slice, this time removing everything after the competition ID. Example: 123
var competitionId = trimmedUrl.Slice(0, competitionIdLength).ToString();

if (!int.TryParse(competitionId, out var parsedCompetitionId)) throw new InvalidOperationException("Unable to extract competition Id from the competition URL");

return parsedCompetitionId;
```

bjss-colin (dismissed):

bjss-colin (approved):

PR #1273, opened 2023-11-01
On-catalogue competition improvements - Download PDFs

DeepikaDK001:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Competitions/Controllers/CompetitionScoringController.cs
```diff
@@ -192,4 +201,84 @@ public class CompetitionScoringController : Controller
 
         return RedirectToAction(nameof(Index), new { internalOrgId, competitionId });
     }
+
+    [HttpGet("featurePdf")]
+    public async Task<IActionResult> FeaturePdf(
```
ghost, line 206: Since all of the following PDF methods are implemented identically, you could instead use a single method and parameterize, in some way, the correct file that needs to be downloaded.

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Competitions/Controllers/CompetitionScoringController.cs
```diff
@@ -192,4 +201,56 @@ public class CompetitionScoringController : Controller
...
+    int competitionId,
+    string pdfType)
+    {
+        var competition = await competitionsService.GetCompetition(internalOrgId, competitionId);
+
+        if (competition == null) return NotFound();
+        string fileName = null;
+        var uri = " ";
+
+        switch (pdfType)
```
ghost: Introducing methods per each non-price element PDF will allow for this switch statement to be removed.

What I meant in my original comment by reusing code is that a single method, that is not `[HttpGet]` decorated, could be used for handling the PDF logic, while other `[HttpGet]` decorated methods pass in the InternalOrgId, CompetitionId, Url and FileName suffix (e.g: features, implementation, interoperability, sla) as parameters.

bjss-colin (changes requested):

imranbashir01 (dismissed):

bjss-colin (dismissed):

bjss-colin (approved):

imranbashir01 (approved):

PR #1286, opened 2023-11-14
Update supported NHS App Integration types

bjss-colin:
Updated the supported NHS App Integration types and I've also refactored the data model.

Where previously the NHS App Integrations were contained in a single `Integration`, this also meant that IM1 and GP Connect integrations would also have an empty `NHSAppIntegrations` collection available to them.

With the revised approach, each Integration represents a single NHS App Integration. This is similar to the current IM1 and GP Connect implementation.


imranbashir01 (dismissed):

src/NHSD.GPIT.BuyingCatalogue.Framework/Constants/Interoperability.cs
```diff
@@ -27,5 +29,33 @@ public static class Interoperability
             { "Appointment Booking", "GP Connect - Appointment Booking" },
             { "Structured Record", "GP Connect - Structured Record" },
         };
+
```
dssherlock, line 33: Why is this information duplicated-isn't the same information available in the Dictionary? Wouldn't that suffice?

dssherlock (dismissed):

src/NHSD.GPIT.BuyingCatalogue.Framework/Constants/Interoperability.cs
ghost: That's true, actually. I added the Descriptions Dictionary after for the view, but I could very easily just enumerate the Dictionary.Keys to fetch what I'm after.

imranbashir01 (approved):

dssherlock (approved):

PR #1327, opened 2024-01-19
Feature/22765 Service Recipients-search modal

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsModalSearch/NhsModalSearch.cshtml
```diff
@@ -0,0 +1,55 @@
+﻿@model NHSD.GPIT.BuyingCatalogue.UI.Components.Views.Shared.Components.NhsModalSearch.NhsModalSearchModel;
+
+<dialog class="modal modal-slide" id="@Model.Id" aria-hidden="true">
+    <form method="dialog">
+        <div class="modal__overlay" tabindex="-1">
+            <div class="modal__container" aria-modal="true" aria-labelledby="@(Model.Id + "-title")">
+                <header class="modal__header">
+                    <h2 class="nhsuk-heading-m nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-0" id="@(Model.Id + "-title")">
```
ghost: How about string interpolation like `@($"{Model.Id}-title")`

As some of these are reused in multiple places, like the `-title` attribute, it might be worth assigning these to a set of variables at the top of the file and then reusing

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsModalSearch/NhsModalSearchViewComponent.cs
```diff
@@ -0,0 +1,34 @@
...
+        private const string DefaultAdvice = "Find and select the Service Recipients you want to include in this order.";
+        private const string DefaultPlaceholder = "Search by ODS code or organisation name";
+        private const string DefaultNotFoundText = "There are no Service Recipients that match. Try entering different search criteria.";
+        private const string DefaultApplyButtonText = "Apply recipients";
+
+        public async Task<IViewComponentResult> InvokeAsync(string id, string showDialogButtonId, string tablePartialView, object tableData, string callbackFunction, string title = null, string advice = null, string placeholder = null, string notFoundText = null, string applyButtonText = null)
+        {
+            var model = new NhsModalSearchModel
+            {
+                Id = id,
```
ghost, line 23: As `Id` is used for HTML attributes, this will need to be sanitized. I can't remember which files in particular do this but there are some examples in this UI Components library for sanitizing IDs.

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsSearchIcon/NhsSearchIcon.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace NHSD.GPIT.BuyingCatalogue.UI.Components.Views.Shared.Components.NhsSearchIcon
+{
+    public sealed class NhsSearchIcon : ViewComponent
+    {
+        public async Task<IViewComponentResult> InvokeAsync()
```
ghost: There's a non async version of Invoke available that would allow for the direct return of `View("NhsSearchIcon")`

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/ServiceRecipients/SelectRecipients.cshtml
```diff
@@ -39,16 +34,23 @@
...
+                    <button id="show-dialog" class="nhsuk-button nhsuk-button--secondary">
+                        <vc:nhs-search-icon/>
+                        Search for Service Recipients
+                    </button>
+                </div>
+            </div>
         }
 
-        <form method="post">
+        <form method="post" id = "sr-form">
```
ghost: HTML formatting is a bit funky on the ID attribute

bjss-colin (changes requested):
Let's exclude ModalSearch.js, ModalSearch.js.map and ModalSearch.min.js as these are artifacts that should be produced at build time when ModalSearch.ts is transpiled into JavaScript

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsModalSearch/NhsModalSearchViewComponent.cs
dssherlock: I'm not absolutely sure what you mean by this. Do u mean use a different term e.g. 'Title' like they do in a few of the components? The only other example i can find using 'Id' is the NhsSuggestionSearch and that doesn't appear to do anything special with it.

ghost: So in the `NhsModalSearch` file the `Id` property is used as the `id` of the parent element. In your example below you assign it an ID of `search-recipients-modal`.

Now what happens if I assign it an ID of the literal `Search Recipients Modal`? That's not a valid value for an id attribute in HTML since it has whitespace between each word, as such it'll need to be sanitized so that `Search Recipients Model` becomes `search-recipients-model` or, a better choice in my opinion, that an exception is thrown since any developer using the component won't know the ID of the element has changed on the basis that their original ID was invalid, meaning the system won't behave as they expect it to.

Sounds like `NhsSuggestionSearch` would need to be revisited as well but we can pick this up separately.

imranbashir01 (dismissed):

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsModalSearch/NhsModalSearchViewComponent.cs
dssherlock: Ah i see yep I'll fix that now

imranbashir01 (approved):

bjss-colin (approved):

PR #1347, opened 2024-02-23
Feature/23404 Manually input Catalogue Item ID

dssherlock:
Using string display parameter to allow binding and prevent default invalid type Error messages. Then validate against CatalogueItemId. Potentially investigate Custom Model Binding as an alternative.


bjss-colin:
Closing for now as this is missing the Additional and Associated Service ID inputs as well as removal the EF Core Value Generator for `CatalogueItemId`

PR #1351, opened 2024-02-29
Feature/23404 add catalogue item ids

dssherlock:


dssherlock:
Closing as change no longer required. Associated service IDs not generated by TechAuth so would still need to be created by the BC. This makes changes for Solution ID and Additional Service IDs redundant.

PR #1353, opened 2024-03-06
23418/bugfix status for additional service

Patrick-M-BJSS:


tests/NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests/Areas/Order/Models/SolutionSelection/TaskList/TaskListModelTests.cs
```diff
@@ -62,7 +63,7 @@ public static class TaskListModelTests
         }
 
         [Theory]
-        [CommonAutoData]
+        [CommonAutoData(MockingFramework.NSubstitute)]
```
ghost, line 66: Let's update all tests in this class to use NSubstitute

tests/NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests/Areas/Order/Models/SolutionSelection/TaskList/TaskListOrderItemModelTests.cs
```diff
@@ -238,85 +239,38 @@ public class TaskListOrderItemModelTests
         }
 
         [Theory]
-        [CommonAutoData]
-        public static void DeliveryDatesStatus_NoOrderItemRecipients_ExpectedResult(
+        [CommonAutoData(MockingFramework.NSubstitute)]
```
ghost, line 242: Same as above - let's update all tests in this class to use NSubstitute

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/AdditionalServices/SelectAdditionalServices.cshtml
```diff
@@ -23,7 +23,7 @@
                         caption="@Model.SolutionName"
                         advice="@advice"/>
         
-        <form method="post">
+        <form method="post" autocomplete="off">
```
ghost, line 26: I'm somewhat curious about this one - what does `autocomplete` do here?

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/AdditionalServices/SelectAdditionalServices.cshtml
Patrick-M-BJSS: This was a minor issue I found while working on it. If you pressed the browser's back button to return to the page, it would remember that the first option (or whichever one you chose) was selected so would automatically select the first etc. one even though none of them should be selected and the options in the list have changed.

ghost: That behaviour should be okay as the user is returning to a page from an existing journey. One thing we'd have to test with this approach is how form validation errors are presented and whether items would be prepopulated in that scenario. It could potentially introduce more edge cases

Patrick-M-BJSS: So if you selected the 3rd element in the list and hit submit, then pressed the back button it would preselect the 3rd element in the list which is now the element that was below the 3rd element because the other one was removed. It's just preselecting random elements which is different behaviour from if you hit the back link. Definitely not correct.

ghost: Ah, understood. The key bit there is that the previously selected service won't be in the list anymore and that a random service would instead be picked.

In that case I'll mark this comment as resolved.

imranbashir01 (approved):

bjss-colin (approved):

PR #1365, opened 2024-03-21
Remove DFOCVC framework and update expiry warning callout

bjss-colin:


imranbashir01 (dismissed):

martinkimmings (dismissed):

src/NHSD.GPIT.BuyingCatalogue.WebApp/wwwroot/ts/FrameworkExpiry.ts
```diff
@@ -0,0 +1,25 @@
+window.addEventListener('load', function () {
+    const warningCallout: HTMLElement = document.getElementById('framework-expired-warning');
+    const container: HTMLElement = warningCallout.querySelector('div[class="container"]');
```
martinkimmings: It's possible that `warningCallout` could be null. The server only renders it when the query string doesn't have any parameters so that, for the non JavaScript version, it disappears when you start filtering.

The JavaScript version for the most part isn't using the query string parameters other than with paging. So I think there's a possibility of getting a JavaScript error here when changing the page after applying some filters.

ghost: That is a good spot. I've added an early return statement if `warningCallout` is null

imranbashir01 (approved):

martinkimmings (approved):

PR #1367, opened 2024-03-22
View all competitions

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/Dashboard/DashboardCardBaseModel.cs
```diff
@@ -8,13 +8,11 @@ public class DashboardCardBaseModel<T>
     public DashboardCardBaseModel(
         string internalOrgId,
         ICollection<T> items,
-        int numberOfItems,
         bool isDashboardView = false,
         PageOptions pageOptions = null)
     {
         InternalOrgId = internalOrgId;
         Items = items;
-        NumberOfItems = numberOfItems;
```
ghost, line 17: `NumberOfItems` was introduced because the Order page presents the total number of orders in the database. With this change, it'll now only show the number of orders based on the search pattern

dssherlock: Ah i couldnt understand where this had come from! Ok I'll remove this change!

ghost: Discussed this on Teams. We should be able to use `PageOptions` to get the desired behaviour

```csharp
public int NumberOfItems => PageOptions?.TotalNumberOfItems ?? Items.Count
```

`PageOptions` will be null on the dashboard view, so `Items.Count` will be used. Conversely, on the detailed pages, `PageOptions` will not be null and the `TotalNumberOfItems` will then achieve the same result as `OrderIds`

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/Dashboard/DashboardCardBaseModel.cs
```diff
@@ -25,9 +23,13 @@ public class DashboardCardBaseModel<T>
 
     public ICollection<T> Items { get; set; }
 
-    public int NumberOfItems { get; set; }
+    public int NumberOfItems => PageOptions?.TotalNumberOfItems ?? Items.Count;
```
ghost, line 26: Don't forget the unit tests for this edge-case 😉 

As I'm off next week, feel free to dismiss my feedback in GitHub to not block the PR. @martinkimmings  should also have permission to force-merge if it becomes an issue.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/Dashboard/DashboardCardBaseModel.cs
```diff
@@ -25,9 +23,13 @@ public class DashboardCardBaseModel<T>
...
 
     public bool ShouldUsePagination => PageOptions is not null;
 
     public bool IsDashboardView { get; set; }
+
+    public bool HasItems => NumberOfItems > 0;
+
+    public int ItemsToIterate => IsDashboardView ? 5 : NumberOfItems;
```
ghost, line 34: This'll need a unit test also to validate the logic

imranbashir01 (approved):

martinkimmings (approved):

PR #1381, opened 2024-04-08
Bugfix/23580 pressing back after deleting shortlist

Patrick-M-BJSS:


imranbashir01 (approved):

bjss-colin (approved):
Nice work on the migration for `ManageFiltersControllerTests`. That class was a bit of a beast

PR #1383, opened 2024-04-11
Prevent user saving a shortlist with zero results

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.WebApp/wwwroot/ts/Filters.ts
```diff
@@ -193,7 +196,7 @@ var filters = (function (): CapabilitiesAndEpics {
 
     function getSortBy(): string {
         const sortOptions = document.getElementById('SelectedSortOption') as HTMLOptionElement;
-        return sortOptions.value;
+        return sortOptions ? sortOptions.value : null;
```
ghost, line 201: As `getSortBy` is used to set query string key-value pairs in the URL, this will result in a `sortBy=null` in the URL. Let's update the code above to prevent that when `getSortBy` returns null

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/wwwroot/ts/Filters.ts
dssherlock: ah yes good spot

bjss-colin:
CodeCov is having a moment again - identifying a change in test coverage where there are existing tests that validate the behaviour.

I'll force-merge this once it's gone through testing.

imranbashir01 (approved):

bjss-colin (approved):

PR #1389, opened 2024-04-18
Notifications: Password Expiry

bjss-colin:
Sends a password expiry notification to active users (not disabled) based on the thresholds of
* 30 days
* 14 days
* 1 day

Uses less than or equal to for the date check so that, at the first deployment, any existing users who are due to expire soon will have emails dispatched.


imranbashir01 (dismissed):

src/BuyingCatalogueFunction/Notifications/ContractExpiry/ContractExpiryFunction.cs
```diff
@@ -41,6 +45,17 @@ public async Task Run([TimerTrigger("0 0 10 * * *")] TimerInfo timerInfo)
         private async Task RunFor(DateTime today)
         {
             logger.LogInformation("Contract Expiry: Evaluating Orders for {Date}", today);
+
+            var defaultEmailPreference = await emailPreferenceService.GetDefaultEmailPreference(EmailPreferenceTypeEnum.ContractExpiry);
+            if (defaultEmailPreference == null)
+            {
+                logger.LogWarning(
+                    "Contract Expiry: {EventToRaise} not found or a ManagedEmailPreference is not configured",
```
martinkimmings: I think this message is a little out of date since the refactor. `EmailPreferenceTypeEnum.ContractExpiry` isn't the event to raise.

src/BuyingCatalogueFunction/Notifications/ContractExpiry/ContractExpiryFunction.cs
```diff
@@ -62,7 +77,7 @@ private async Task RunFor(DateTime today)
             }
         }
 
-        private async Task EvaluateOrder(DateTime today, Order order)
+        private async Task EvaluateOrder(DateTime today, Order order, EmailPreferenceType defaultEmailPreference)
         {
             var eventToRaise = order.EndDate.DetermineEventToRaise(today, order.ContractOrderNumber.OrderEvents);
             if (eventToRaise == EventTypeEnum.Nothing)
```
martinkimmings: Should we check that the `EventToRaise` is covered by the `ContractExpiry` email preference somewhere. This was initially checked when the email preference was loaded from the event via the `EventTypes` table. The refactoring makes sense in that the `ContractExpiryFunction` is only interested in the `ContractExpiry` email preference.

src/BuyingCatalogueFunction/Notifications/ContractExpiry/ContractExpiryFunction.cs
```diff
@@ -70,26 +77,18 @@ private async Task RunFor(DateTime today)
             }
         }
 
-        private async Task EvaluateOrder(DateTime today, Order order)
+        private async Task EvaluateOrder(DateTime today, Order order, EmailPreferenceType defaultEmailPreference)
         {
             var eventToRaise = order.EndDate.DetermineEventToRaise(today, order.ContractOrderNumber.OrderEvents);
```
martinkimmings, line 82: Should we check that the `EventToRaise` is covered by the `ContractExpiry` email preference somewhere. This was initially checked when the email preference was loaded, from the event, via the `EventTypes` table. The refactoring makes sense in that we expect the `ContractExpiryFunction` is only interested in the `ContractExpiry` email preference but it opens up the question are we using the right email preference for the event we are about to raise.

src/BuyingCatalogueFunction/Notifications/ContractExpiry/ContractExpiryFunction.cs
ghost: Renamed to `EmailPreferenceType`

src/BuyingCatalogueFunction/Notifications/ContractExpiry/ContractExpiryFunction.cs
ghost: Added a navigation from `EmailPreferenceType` to `EventType` in order to validate the event to raise against the notification preference

martinkimmings (approved):

imranbashir01 (approved):

PR #1397, opened 2024-05-01
Feature/23520 Pagination Component

dssherlock:
Add the govuk-frontend pagination component in the following places:

- Catalogue Solutions
- View Orders
- View Competitions


src/NHSD.GPIT.BuyingCatalogue.Services/Competitions/CompetitionsService.cs
```diff
@@ -36,12 +38,30 @@ await dbContext.Competitions
...
         => await dbContext.Competitions.Include(x => x.CompetitionSolutions)
             .Where(x => x.Organisation.InternalIdentifier == internalOrgId)
             .IgnoreQueryFilters()
             .ToListAsync();
 
+    public async Task<PagedList<Competition>> GetPagedCompetitions(string internalOrgId, PageOptions options)
+    {
+        options ??= new PageOptions();
+
+        var query = await GetCompetitions(internalOrgId);
```
ghost: This will be a list rather than a query at this point, since `GetCompetitions` uses `ToListAsync`. Since this is dynamically sized, it might be better to craft a query here so that the pagination will take effect at a database level.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/SearchResults.cshtml
```diff
@@ -69,15 +69,12 @@
         }
     }
     
-    <nhs-page-number-pagination
-        next-url="@(Url.Action(
+    <govuk-pagination
+        base-url="@(Url.Action(
```
ghost, line 73: Thinking about the recent pentest and the issue where embedded URLs were called out as a risk for cross-site-scripting, I'm wondering if it's possible to achieve this without specifying a full URL.

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/TagHelpers/GDSPagination/GDSPaginationTagHelper.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                var builder = new UriBuilder(url);
+                var query = HttpUtility.ParseQueryString(builder.Query);
+                query.Set("page", pageNumber.ToString());
+                builder.Query = query.ToString() ?? string.Empty;
+                return builder.Uri.ToString();
+            }
+
+            var url = string.IsNullOrWhiteSpace(BaseUrl)
+                ? ViewContext.HttpContext.Request.GetEncodedUrl()
```
ghost, line 120: Further to my comment on the view about XSS, `GetEncodedUrl` seems to work sufficiently here when I've tested this locally. Each time the JavaScript variant of filtering is used, `GetEncodedUrl` returns the correct URL.

The issue with `BaseUrl` in this case is that we're using `AppendHtml` to embed that in the body. This completes the attack vector since the user browser will treat malicious HTML, injected via the query string, as valid HTML.

One thing I have spotted is that `GetEncodedUrl` maybe returns a bit _too_ much info when using the JavaScript based filtering process. The TypeScript may need to be revisited so that it's only setting query string params when the value is not null.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/SearchResults.cshtml
```diff
@@ -69,15 +69,12 @@
         }
     }
     
-    <nhs-page-number-pagination
-        next-url="@(Url.Action(
+    <govuk-pagination
```
ghost, line 72: It's entirely possible that I'm being dense, but having done an `npx gulp min` and a rebuild, I'm not seeing any styling on the pagination component

image

bjss-colin (changes requested):
Some minor points and one point related to the recent pentest

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/TagHelpers/GDSPagination/GDSPaginationTagHelper.cs
dssherlock: No ive definitely had issues when it just uses GetEncodedUrl as it returns the searchResult ajax callback url instead.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/SearchResults.cshtml
dssherlock: So i think the reason this was used in the original pagination component in the first place is because when you are on the Solutions page, the urls are initially all set correctly, however as soon as you start filtering, it does an ajax callback to update the search results. If you don't use that base url and just use the Request Url, then the new page urls will use the ajax callback instead and when the user clicks on it, they just see a partial view rendered on the page.

dssherlock: I did half consider in this case trying to use the Request.RefererUrl instead as this could possibly work, but decided to go with the existing logic as it reduced the testing required.

dssherlock: Another issue i found with this is that the HttpUtility library does not seem to work well with relative urls. So any of the standard ways of updating a parameter using ParseQueryString become a bit unwieldy unless u use the Url.ActionContext.HttpContext.Request.Scheme. I guess always possible to just use a regex to update the page parameter but its not encouraged.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/SearchResults.cshtml
dssherlock: Tbh i messed around with my local environment so many times swapping between current, version 4.2 and the latest version im not 100% sure how i ended up with it working!
image
I think this is what got it working in the end.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/SearchResults.cshtml
ghost: Fair point - I'd tested based on some of the quick filters available which happened to match the edge cases where `GetEncodedUrl` works. It works all of the time when not using JavaScript since the app uses GET and POST.

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/TagHelpers/GDSPagination/GDSPaginationTagHelper.cs
ghost: Conversation resolved based on previous point

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/SearchResults.cshtml
ghost: That's done the trick locally. This change will need persisting in the PR so the build agent can pick up the relevant versions and artifacts

src/NHSD.GPIT.BuyingCatalogue.WebApp/package-lock.json
```diff
@@ -2524,6 +2526,7 @@
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/glogg/-/glogg-1.0.2.tgz",
       "integrity": "sha512-5mwUoSuBk44Y4EshyiqcH95ZntbDdTQqA3QYSrxmzj28Ai0vXBGMH1ApSANH14j2sIRtqCEyg6PfsuP7ElOEDA==",
+      "dev": true,
```
dssherlock, line 2529: I'm not sure these should be included tho? They were added when did the install

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Solutions/SearchResults.cshtml
dssherlock: I'm not sure how this will work as part of the build and release process tho. I assumed updating the version in package.json and package-lock would be sufficient. There are some other changes that appeared in the package-lock.json file that I'm not convinced should be there though-the additional "dev" property, along with a few bits that have been removed. This happened as part of the npm install process. Thoughts?

ghost: > I'm not sure how this will work as part of the build and release process th

It's a standard package restore process much like NuGet is. Packages will be restored at build time based on the packages and versions listed in package.json. package-lock.json just freezes the dependencies so that others are able to install the same dependencies in the same manner.

> There are some other changes that appeared in the package-lock.json file that I'm not convinced should be there though-the additional "dev" property

The `dev` property isn't referring to any of the dependencies required to run this in a release mode. They're packages required specifically for development processes, such as the gulp build process. This can/should be tested locally by using the `--production` switch with `npm install`.

In this case, the govuk-frontend package version needs to be updated to `4.6.0` otherwise the build agent won't restore the correct version and subsequently the styles won't work on any environment.

dssherlock: Ok yep thats what i thought. So i will just leave the legitimate version update changes in the package-lock.json file and revert the other ones (in red)
image

ghost: I'd leave the package-lock.json as it currently is since this file is generated automatically. Manual changes will be overriden the next time this is built and a pull request raised. We literally just need to update the package version of govuk-frontend to 4.6.0 (`npm install govuk-frontend@4.6.0`) and that'll resolve this one

dssherlock: Ok why 4.6 though? I thought that was a typo!!

imranbashir01 (approved):

bjss-colin (approved):

PR #1400, opened 2024-05-07
Regression tests add solution

imranbashir01:


tests/NHSD.GPIT.BuyingCatalogue.RegressionTests/Pages/Admin/ManageSolutions/SolutionApplicationType/BrowserBased.cs
```diff
@@ -0,0 +1,168 @@
...
+            CommonActions.ClickLinkElement(AddApplicationTypeObjects.PluginsOrExtensionsLink(value));
+            CommonActions.PageLoadedCorrectGetIndex(
+                typeof(BrowserBasedController),
+                nameof(BrowserBasedController.PlugInsOrExtensions))
+                .Should().BeTrue();
+
+            CommonActions.ClickFirstRadio();
+            TextGenerators.TextInputAddText(AddApplicationTypeObjects.AdditionalInformation, 500);
+            CommonActions.ClickSave();
+
```
dssherlock, line 96: Would it possibly be an idea to check the status has changed here? Not sure of the flow of the tests so this may not be the right place to do it. Obviously a single function that would be called after adding each browser based application type.

dssherlock (approved):

tests/NHSD.GPIT.BuyingCatalogue.RegressionTests/Pages/Admin/ManageSolutions/SolutionApplicationType/BrowserBased.cs
imranbashir01: it would be difficult to tag the status against the application types based as it's not tagged against any application type, so let's say if first one is completed it will always be true for the rest
image

bjss-colin (approved):

PR #1399, opened 2024-05-07
Order triage and form framework changes

martinkimmings:

PR #1401, opened 2024-05-08
IA shortlist changes

Patrick-M-BJSS:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/Shortlists/_ResultsForFramework.cshtml
```diff
@@ -0,0 +1,44 @@
...
+
+@if (Model.CatalogueItems?.Any() ?? false)
+{
+    <hr class="nhsuk-u-padding-bottom-5 nhsuk-u-margin-bottom-5 nhsuk-u-margin-top-0">
+    <div>
+        <h2 class="nhsuk-heading-s nhsuk-u-margin-bottom-4">Results from @Model.Framework.ShortName (@Model.CatalogueItems.Count())</h2>
+        <p class="nhsuk-body-s nhsuk-u-margin-top-3">You can only carry out competitions using solutions from the same framework. The results are grouped by framework so you can create a valid competition. </p>
+        <p><b>All links open in a new tab.</b></p>
+    </div>
+    <h3></h3>
```
ghost: Empty H3

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Competitions/Views/CompetitionSelectSolutions/_MultiSolutionSelection.cshtml
```diff
@@ -14,32 +10,60 @@
...
-                <nhs-checkbox asp-for="@Model.Solutions[i].Selected"
-                              label-text="<a href='@GetSolutionMarketingPageUrl(Model.Solutions[i].SolutionId)' target='_blank'>@Model.Solutions[i].SolutionName (opens in a new tab)</a>"
-                              hint-text="@Model.Solutions[i].SupplierName"
-                              sub-hint="@content @Model.Solutions[i].GetAdditionalServicesList()"
-                              use-embedded-html="true"/>
+                <input type="hidden" asp-for="@Model.Solutions[i].RequiredServices[x]"/>
             }
-        </nhs-checkbox-container>
+
+            <nhs-card-v2 class="nhsuk-card__inverted" style="margin-bottom:20px">
```
ghost: I imagine we'll want to use the inverted cards elsewhere.

This could be a property on the tag helpper, such as `IsInverted` (HTML attribute name `inverted`), that conditionally adds the `nhsuk-card__inverted` class when set to true via html attributes. There are some examples in this PR where the inverted card is used in multiple places, that could be swapped out with an attribute.

Example:

```html
<nhs-card-v2 inverted="true"></nhs-card-v2>
```

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/SolutionsFilterService.cs
```diff
@@ -165,6 +165,22 @@ public async
             return (results, options, count);
         }
 
+        public async
+            Task<(IList<CatalogueItem> CatalogueItems, PageOptions Options, List<CapabilitiesAndCountModel>
```
ghost: Doesn't look like elements 2 & 3 in this tuple are used anywhere. This method return could be simplified

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Competitions/Controllers/CompetitionsDashboardController.cs
```diff
@@ -126,19 +147,37 @@ public IActionResult ReviewFilter(string internalOrgId, int filterId, ReviewFilt
...
-            BackLink = fromFilter == true ?
+        var backlink = fromFilter == true ?
                 Url.Action(
                     nameof(ManageFiltersController.FilterDetails),
                     typeof(ManageFiltersController).ControllerName(),
                     new { filterId, Area = typeof(ManageFiltersController).AreaName() }) :
-                Url.Action(nameof(ReviewFilter), new { internalOrgId, filterId }),
+                Url.Action(nameof(ReviewFilter), new { internalOrgId, filterId });
+
+        frameworkId = (await filterService.GetFilterIds(organisation.Id, filterId))?.FrameworkId ?? frameworkId;
```
ghost: Logic seems to be the wrong way around here since the value in `frameworkId` should take precendence.

This whole check could be removed since the `frameworkId` is passed in via the shortlist results page. Rather than defaulting, if the value is manipulated in the URL then we should return.

The two things to validate
1. `frameworkId` is not null or empty (prefer `string.IsNullOrWhitespace` rather than `framework.IsNullOrEmpty`)
2. Is it a valid framework that exists in the database. Query the `IFrameworkService.GetFramework` and check for null

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/ManageFiltersController.cs
```diff
@@ -169,9 +175,14 @@ public async Task<IActionResult> FilterDetails(int filterId)
                 BackLink = Url.Action(nameof(Index), typeof(ManageFiltersController).ControllerName()),
                 Caption = organisation.Name,
                 InternalOrgId = organisation.InternalIdentifier,
+                FilterResults = (List<EntityFramework.Catalogue.Models.CatalogueItem>)solutions,
```
ghost: `.ToList()` should work here rather than casting, since casts could fail

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/ManageFiltersController.cs
```diff
@@ -160,6 +164,8 @@ public async Task<IActionResult> FilterDetails(int filterId)
             var organisation = await GetUserOrganisation();
             var filterDetails = await manageFiltersService.GetFilterDetails(organisation.Id, filterId);
             var filterIds = await manageFiltersService.GetFilterIds(organisation.Id, filterId);
+            var (solutions, _, _) = await solutionsFilterService.GetAllSolutionsFilteredFromFilterIds(filterIds);
+            var frameworks = await frameworkService.GetFrameworksWithPublishedCatalogueItems();
```
ghost: Worth sanity checking this logic with `frameworks` to think about this approach. If I save a shortlist that includes TIF as a framework, I wouldn't necessarily expect to see DFOCVC in the shortlist results page :wink:

image

image


The frameworks should be driven from the solutions returned from the shortlist. The model should manipulate that solutions list (hint: `GroupBy`) to group the solutions by framework. Bear in mind that models are great places for massaging data into the required format, rather than doing too much in the Controller or Service layers.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/Shortlists/FilterDetails.cshtml
```diff
@@ -0,0 +1,84 @@
...
+        <nhs-expander label-text="View filters used for this shortlist" open="false">
+            <p>These are the filters used in this shortlist:</p>
+
+            <partial name="Partials/_FilterDetailsPartial" model="Model" />
+        </nhs-expander>
+            
+        @if(hasResults && !Model.InCompetition){
+            <vc:nhs-delete-button url="@Url.Action(nameof(ManageFiltersController.DeleteFilter),
+                                                        new { filterId = Model.FilterDetails.Id })"
+                                  text="Delete filter" />
```
ghost: Should this read `Delete shortlist` instead?

bjss-colin (changes requested):
Rough first review pass: spotted an issue with filtering logic and a few minor code cleanup points

bjss-colin (changes requested):
As Martin's PR has been merged (see #1399), let's also update `CompetitionOrderService.CreateOrder` to pass in the Framework from the Competition, and `CompetitionSelectSolutionsController.HandleDirectAward` so we're passing the framework in via `selectedFrameworkId`

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ReviewFilterModel.cs
```diff
@@ -16,12 +19,46 @@ public ReviewFilterModel()
         FilterIds = filterIds;
     }
 
+    public ReviewFilterModel(FilterDetailsModel filterDetails, string internalOrgId, List<FrameworkFilterInfo> frameworks, List<CatalogueItem> filterResults, bool inCompetition, FilterIdsModel filterIds = null)
```
ghost: How about this?

Note that `frameworks` isn't actually needed. As mentioned above, that data is already discoverable via `CatalogueItem.Solution.FrameworkSolutions`. `CatalogueItem` is the type held within the `filterResults` collection.

Also worth noting that `ResultsForFrameworkModel` has been updated since that only needs access to the Framework Id and Framework Name.

It'll be good to sanity check the data returned in various filters such as framework only, capabilities and epics etc.

```csharp
public ReviewFilterModel(FilterDetailsModel filterDetails, string internalOrgId, List<CatalogueItem> filterResults, bool inCompetition, FilterIdsModel filterIds = null)
    : this(filterDetails, filterIds)
{
    ArgumentNullException.ThrowIfNull(filterIds);

    ResultsCount = filterResults.Count;
    InCompetition = inCompetition;

    var selectedFrameworks = filterResults
        .SelectMany(x => x.Solution.FrameworkSolutions)
        .GroupBy(x => (x.Framework.ShortName, x.Framework.Id));

    if (!string.IsNullOrWhiteSpace(filterIds.FrameworkId))
        selectedFrameworks = selectedFrameworks.Where(x => string.Equals(x.Key.Id, filterIds.FrameworkId));

    ResultsForFrameworks = selectedFrameworks.Select(
            x => new ResultsForFrameworkModel(
                internalOrgId,
                filterDetails.Id,
                x.Key.Id,
                x.Key.ShortName,
                x.Select(y => y.Solution.CatalogueItem).ToList(),
                !inCompetition))
        .ToList();
}
```

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/ManageFiltersController.cs
ghost: Reiterating that `frameworks` isn't actually required here. See comment below on the model for more

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/Shortlists/FilterDetails.cshtml
```diff
@@ -0,0 +1,83 @@
...
+                                Delete this shortlist
+                            </a>
+                        </div>
+                        
+                    }
+                    else
+                    {
+                        foreach (var resultsForFrameworkModel in Model.ResultsForFrameworks)
+                        {
+                            <partial name="_ResultsForFramework.cshtml" model="resultsForFrameworkModel"/>
```
ghost: .cshtml not required here

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Competitions/Controllers/CompetitionsDashboardController.cs
```diff
@@ -104,17 +116,22 @@ public async Task<IActionResult> ReviewFilter(string internalOrgId, int filterId
...
-            BackLink = Url.Action(nameof(SelectFilter), new { internalOrgId }),
-            Caption = filterDetails.Name,
+            BackLink = Url.Action(nameof(Index), typeof(ManageFiltersController).ControllerName()),
+            Caption = organisation.Name,
+            OrganisationName = organisation.Name,
+            InExpander = true,
         };
 
-        return View(model);
+        return View("Views/Shared/Shortlists/FilterDetails.cshtml", model);
```
ghost: `View` takes a view name rather than a file path. `.cshtml` not required here and `Views/Shared` can be removed 😃 

Worth doing a Find All to find similar instances (should be 2 in this PR)

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/Shortlists/FilterDetails.cshtml
```diff
@@ -0,0 +1,83 @@
...
+@using NHSD.GPIT.BuyingCatalogue.Framework.Extensions
+@using NHSD.GPIT.BuyingCatalogue.WebApp.Areas.Solutions.Models.ManageFilters
+@using NHSD.GPIT.BuyingCatalogue.WebApp.Models.Shared.Shorlists
+
+@model NHSD.GPIT.BuyingCatalogue.WebApp.Models.Shared.ReviewFilterModel
+
+@{
+    Layout = "~/Views/Shared/Layouts/_AllBannersLayout.cshtml";
+    ViewBag.Title = Model.FilterDetails.Name;
+    var hasResults = @Model.ResultsCount > 0;
```
ghost: `@` not required here since this is already within the context of a code block

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/Shortlists/FilterDetails.cshtml
```diff
@@ -0,0 +1,83 @@
...
+                    {
+                        <hr class="nhsuk-u-padding-bottom-5 nhsuk-u-margin-bottom-5 nhsuk-u-margin-top-0">
+                        <p>Due to changes on the Buying Catalogue, there are currently no results returned using the selected filters. This may change in future as a wider range of solutions are added.</p>
+                        <div style="display: inline-block" class="nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-4">
+                            <vc:nhs-secondary-button url="@Url.Action(
+                                      nameof(ManageFiltersController.Index),
+                                      typeof(ManageFiltersController).ControllerName())"
+                                                     type="Secondary"
+                                                     text="Return to shortlists" />
+                            <a href="@Url.Action(nameof(ManageFiltersController.DeleteFilter),
```
ghost: There is an existing component for delete links - `vc:nhs-delete-button`

That could be wrapped in a `<span></span>` which then applies the additional styles :wink:

bjss-colin (changes requested):
Some additional code cleanup suggestions. Please do resolve conversations in GitHub when fixes have been applied

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/Shortlists/FilterDetails.cshtml
Patrick-M-BJSS: It only seems to work when there is a .cshtml

ghost: What happens in that scenario? What error occurs and what paths does the error contain? 

Hint: neither `Views` nor `Shared` are required in the name, but the name of a certain subdirectory is required.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/ManageFiltersController.cs
```diff
@@ -160,18 +164,20 @@ public async Task<IActionResult> FilterDetails(int filterId)
...
             {
                 BackLink = Url.Action(nameof(Index), typeof(ManageFiltersController).ControllerName()),
                 Caption = organisation.Name,
-                InternalOrgId = organisation.InternalIdentifier,
+                OrganisationName = organisation.Name,
+                InExpander = true,
             };
 
-            return View(model);
+            return View("Views/Shared/Shortlists/FilterDetails.cshtml", model);
```
ghost: Looks like this was missed during the recent rebase.

View path can be changed to `Shortlists/FilterDetails` without the cshtml and without the `Views/Shared` path-part. ASP.NET Core already knows to look in `Views/Shared` for shared views.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/Shortlists/FilterDetails.cshtml
```diff
@@ -0,0 +1,85 @@
...
+                        <hr class="nhsuk-u-padding-bottom-5 nhsuk-u-margin-bottom-5 nhsuk-u-margin-top-0">
+                        <p>Due to changes on the Buying Catalogue, there are currently no results returned using the selected filters. This may change in future as a wider range of solutions are added.</p>
+                        <div style="display: inline-block" class="nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-4">
+                            <vc:nhs-secondary-button url="@Model.BackLink"
+                                                     type="Secondary"
+                                                     text="Return to shortlists" />
+                            @if (!Model.InCompetition)
+                            {
+                                <span class="modal-link-adjuster nhsuk-u-margin-left-3" draggable="false">
+                                    <vc:nhs-delete-button url="@Url.Action(nameof(ManageFiltersController.DeleteFilter),
```
ghost, line 52: This link looks like it'll break on the Competition journey as we saw during the demo. Let's fix this.

When building links in ASP.NET core, the Action, Controller and Area are all mandatory if the user is being redirected to a different controller and area.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/Shortlists/FilterDetails.cshtml
```diff
@@ -0,0 +1,85 @@
...
+        
+
+        <nhs-expander label-text="View filters used for this shortlist" open="false">
+            <p>These are the filters used in this shortlist:</p>
+
+            <partial name="Partials/_FilterDetailsPartial" model="Model" />
+        </nhs-expander>
+            
+        @if(hasResults && !Model.InCompetition){
+            <vc:nhs-delete-button url="@Url.Action(nameof(ManageFiltersController.DeleteFilter),
```
ghost, line 81: Same as above, this link isn't specific enough that would allow ASP.NET Core to choose the correct link, since it isn't being told which Controller and Area the `DeleteFilter` action lives on.

bjss-colin (changes requested):
1 minor point and 2 around the link issue we saw during the demo on Friday

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Competitions/Views/CompetitionSelectSolutions/_MultiSolutionSelection.cshtml
```diff
@@ -2,44 +2,94 @@
...
     <input type="hidden" asp-for="BackLink"/>
     <input type="hidden" asp-for="BackLinkText"/>
     <input type="hidden" asp-for="CompetitionName"/>
+    <input type="hidden" asp-for="FrameworkName"/>
+    <input type="hidden" asp-for="ReviewFilter.FilterDetails.Capabilities"/>
+    <input type="hidden" asp-for="ReviewFilter.FilterDetails.FrameworkName"/>
+    <input type="hidden" asp-for="ReviewFilter.InExpander"/>
+    @for (int i = 0; i < Model.ReviewFilter.FilterDetails.ApplicationTypes.Count; i++)
+    {
+        @Html.HiddenFor(m => m.ReviewFilter.FilterDetails.ApplicationTypes[i])
```
ghost: Inconsistent approaches: let's use the `<input type="hidden" asp-for="..." />` approach as above

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Competitions/Views/CompetitionSelectSolutions/_SingleSolutionSelection.cshtml
```diff
@@ -25,9 +25,35 @@
...
+    <input type="hidden" asp-for="ReviewFilter.FilterDetails.FrameworkName" />
+    <input type="hidden" asp-for="ReviewFilter.InExpander" />
+    <input type="hidden" asp-for="@Model.Solutions[0].SolutionId" />
+    <input type="hidden" asp-for="@Model.Solutions[0].SolutionName" />
+    <input type="hidden" asp-for="@Model.Solutions[0].SupplierName" />
+    <input type="hidden" asp-for="@Model.Solutions[0].RequiredServices" />
+    <input type="hidden" asp-for="@Model.Solutions[0].Summary" />
+    @for (int i = 0; i < Model.ReviewFilter.FilterDetails.ApplicationTypes.Count; i++)
+    {
+        @Html.HiddenFor(m => m.ReviewFilter.FilterDetails.ApplicationTypes[i])
```
ghost: Inconsistent approaches: let's use the `<input type="hidden" asp-for="..." />` approach as above

bjss-colin (changes requested):
Minor points from latest commit

imranbashir01 (changes requested):
links on the Results for the shortlist page are broken, I get an incorrect URL page when I click on the link. Also a full stop missing from "All links open in a new tab"

imranbashir01 (approved):

bjss-colin (approved):

PR #1435, opened 2024-06-19
Feature/23321 nhs app filters

dssherlock:


imranbashir01 (dismissed):

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Configuration/InteropNhsAppIntegrationType.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace NHSD.GPIT.BuyingCatalogue.EntityFramework.Catalogue.Configuration;
+
+public enum InteropNhsAppIntegrationType
+{
+    [Display(Name = "Online Consultations")]
+    Online_Consultations,
```
ghost: Let's add explicit integers here. Enums apply ints based on the order so if, for example, `Online_Consultations` is moved to the end of this enum then the implicit int will change from 0 to 3 and all other enum values will decrement by -1.

bjss-colin (changes requested):
One minor point

bjss-colin (approved):

PR #1438, opened 2024-06-27
23637 admin manage orders

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderAdminService.cs
```diff
@@ -59,31 +62,53 @@ public async Task<Order> GetOrder(CallOffId callOffId)
...
+            if (!string.IsNullOrWhiteSpace(framework))
+            {
+                baseQuery = baseQuery.Where(
+                    o => o.SelectedFrameworkId.Contains(framework));
+            }
+
+            var orderList = await baseQuery.ToListAsync();
+            if (!string.IsNullOrWhiteSpace(status))
+            {
+                if (Enum.TryParse(status, true, out OrderStatus orderStatus)
```
ghost: ASP.NET Core will handle this out of the box for us. Pass `OrderStatus?` to the controller query string parameters and leave model binding handle it.

We did similar with the older `OrderTriageValue` enum that was passed around in the order journey, though that's now removed.

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderAdminService.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
```
ghost: I don't think these 3 using statements are required

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/ManageOrders/ManageOrdersDashboardModel.cs
```diff
@@ -1,21 +1,77 @@
...
+        public string SelectedFramework { get; set; }
+
+        public List<FrameworkFilterInfo> Frameworks { get; set; }
+
+        public IEnumerable<SelectOption<string>> AvailableFrameworks { get; set; }
+
+        public OrderStatus? SelectedStatus { get; set; }
+
+        public IEnumerable<SelectOption<OrderStatus>> AvailableStatus =>
+            Enum.GetValues(typeof(OrderStatus))
```
ghost: Could use `Enum.GetValues<OrderStatus>()` here which would remove the need for `Cast<OrderStatus>()`

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Controllers/ManageOrdersController.cs
```diff
@@ -25,47 +34,65 @@ public class ManageOrdersController : Controller
...
                 BackLink = Url.Action(nameof(HomeController.Index), typeof(HomeController).ControllerName()),
             };
 
             return View(model);
         }
 
+        [HttpPost]
+        public IActionResult Index(ManageOrdersDashboardModel model)
+        {
+            return RedirectToAction(
```
ghost, line 72: What happens if the form is submitted with no selection?

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/ManageOrders/ManageOrdersDashboardModel.cs
```diff
@@ -1,21 +1,77 @@
...
 
         public IList<AdminManageOrder> Orders { get; set; }
 
         public PageOptions Options { get; set; }
+
+        public int FilterCount { get; set; }
+
+        public string SelectedFramework { get; set; }
+
+        public List<FrameworkFilterInfo> Frameworks { get; set; }
```
ghost: Doesn't look like this is used anywhere

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderAdminService.cs
```diff
@@ -59,31 +62,53 @@ public async Task<Order> GetOrder(CallOffId callOffId)
...
                 baseQuery = GetSearchTermBySearchType(baseQuery, search, searchTermType);
 
-            options.TotalNumberOfItems = await baseQuery.CountAsync();
+            if (!string.IsNullOrWhiteSpace(framework))
+            {
+                baseQuery = baseQuery.Where(
+                    o => o.SelectedFrameworkId.Contains(framework));
+            }
+
+            var orderList = await baseQuery.ToListAsync();
```
ghost, line 84: This will pull in all orders, from the database, not matching the order status and apply filtering in-memory

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Controllers/ManageOrdersController.cs
```diff
@@ -25,47 +34,65 @@ public class ManageOrdersController : Controller
...
         }
 
         [HttpGet]
         public async Task<IActionResult> Index(
             [FromQuery] string page = "",
             [FromQuery] string search = "",
-            [FromQuery] string searchTermType = "")
+            [FromQuery] string searchTermType = "",
+            [FromQuery] string framework = "",
+            [FromQuery] string status = "")
```
ghost: This can be `OrderStatus?` with a default of `null`

bjss-colin (changes requested):
Some minor model binding observations and potentially unused properties

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderAdminService.cs
dssherlock: Yes i didn't believe we could filter on order status as this isn't an actual value in the DB-it is a property that is calculated in-memory.

ghost: I'll add a cleanup activity to remove the `OrderStatusId` column, and other unused columns, in the database

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Controllers/ManageOrdersController.cs
dssherlock: Technically nothing-it will just redirect to the index with the filters set to nothing-so it will basically act like the clear all filters link. But it could leave the url looking a bit odd i guess (status=&framework=.

ghost: I don't think we need to add validation or anything but it might not be a bad idea to set the values only if they're populated

dssherlock: Actually having debugged it, it appears to be clever enough to only add the route values if they are not null. So it is all fine.

ghost: Ah, even better. That's perfect

src/NHSD.GPIT.BuyingCatalogue.Services/Framework/FrameworkService.cs
```diff
@@ -33,6 +33,16 @@ await dbContext.FrameworkSolutions.AsNoTracking()
                     })
                 .ToListAsync();
 
+        public async Task<List<FrameworkFilterInfo>> GetAllFrameworks() =>
```
ghost: This does the same thing as `GetFrameworks` in this class, but with a different model. Let's update the code to use the existing `GetFrameworks`

bjss-colin (changes requested):

imranbashir01 (dismissed):

imranbashir01 (approved):

bjss-colin (approved):

PR #1441, opened 2024-07-01
Add Integrations to database

bjss-colin:
Deprecates the JSON based Integrations column in the Solutions table by moving the Integrations and Integration Types to their own database tables.

Implements the following schema changes on Solutions, Competitions and Filtering tables.

Solutions:
image

Filtering:
image

Competitions
image

This change also removes the Integration Tests suite as this hasn't been used in over a year due to performance bottlenecks with the build agents. Additionally, I've done a solution wide `using` cleanup. These two changes contribute to a lot of the 760+ files changed.


database/NHSD.GPITBuyingCatalogue.Database/PostDeployment/Seed/InsertSeedIntegrations.sql
```diff
@@ -0,0 +1,27 @@
+﻿IF NOT EXISTS(SELECT 1 FROM catalogue.Integrations)
```
dssherlock: Wouldn't it be wise to separate the seeding of these two tables? This feels right for the Integration type table as it is a one-off seeding and then it becomes user controlled. But I wonder if the Integrations table should really use the merge functionality as for the time being this is essentially reference data?

ghost: That's a good point - done

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Catalogue/Models/SupportedIntegrations.cs
```diff
@@ -1,13 +1,13 @@
 ﻿using System.Runtime.Serialization;
```
dssherlock, line 2: I'm still not convinced i understand why this Enum is necessary when you have the Integration table. Why can't we just use an IntegrationId like we do elsewhere. I'm putting this here to remind myself to come back to it when playing around in the code!

ghost: The enum is mainly to avoid using magic numbers in code by providing description/meaning to the IDs since these are used on the admin side when an admin is adding integration types for a specific integration. This is currently handled in many Integration-specific actions on the `InteroperabilityController` which use the `GetIntegrationTypesByIntegration` method of `IIntegrationsService`

Although I'd considered refactoring that controller, it would have made an already large pull-request even larger.

dssherlock: Ah ok that makes sense. I thought maybe I'd missed something rather than it just being a pragmatic approach.

src/NHSD.GPIT.BuyingCatalogue.Services/ServiceHelpers/SolutionsFilterHelper.cs
```diff
@@ -29,41 +27,26 @@ public static class SolutionsFilterHelper
...
+                .Select(x => x.Split(FilterConstants.Delimiter, StringSplitOptions.RemoveEmptyEntries & StringSplitOptions.TrimEntries))
+                .Where(x => Enum.TryParse(typeof(SupportedIntegrations), x[0], out var parsed) && Enum.IsDefined(typeof(SupportedIntegrations), parsed))
+                .Select(x => new KeyValuePair<SupportedIntegrations, int[]>(Enum.Parse<SupportedIntegrations>(x[0]), x.Skip(1).Where(y => int.TryParse(y, out _)).Select(int.Parse).ToArray())));
+        }
 
-        public static ICollection<InteropIntegrationType> ParseInteropIntegrationTypeIds(string selectedInteroperabilityIds) =>
-            selectedInteroperabilityIds?.Split(FilterConstants.Delimiter, StringSplitOptions.RemoveEmptyEntries & StringSplitOptions.TrimEntries)
-                .Where(x => Enum.TryParse(typeof(InteropIntegrationType), x, out var hostingValue) && Enum.IsDefined(typeof(InteropIntegrationType), hostingValue))
-                .Select(t => (InteropIntegrationType)Enum.Parse(typeof(InteropIntegrationType), t))
-                .ToList() ?? new List<InteropIntegrationType>();
```
dssherlock, line 66: Maybe make argument name more generic (not hostingTypeIds) as used in multiple places

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/InteroperabilityService.cs
```diff
@@ -29,101 +21,67 @@ public async Task SaveIntegrationLink(CatalogueItemId solutionId, string integra
...
+            var integration = solution.Integrations.FirstOrDefault(x => x.Id == integrationId);
+            if (integration is null) return;
 
-                solution.Integrations = JsonSerializer.Serialize(integrations);
-
-                await dbContext.SaveChangesAsync();
-            }
+            solution.Integrations.Remove(integration);
+            await dbContext.SaveChangesAsync();
         }
```
dssherlock, line 61: Trying to get to the bottom of why adding a new NhsApp Integration requires its own special function (why can't it just use interoperabilityService.AddIntegration like GP connect or IM1)? 

N.B. This is one for me to figure out Colin rather than anything you have done!

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/InteroperabilityModel.cs
```diff
@@ -15,34 +14,22 @@ public InteroperabilityModel()
         public InteroperabilityModel(CatalogueItem catalogueItem, CatalogueItemContentStatus contentStatus)
             : base(catalogueItem, contentStatus)
         {
-            if (catalogueItem is null)
-                throw new ArgumentNullException(nameof(catalogueItem));
+            ArgumentNullException.ThrowIfNull(catalogueItem);
 
-            var integrations = catalogueItem.Solution?.GetIntegrations();
```
dssherlock, line 21: Why remove the null check for a Solution? What if the Cat Item is an Associated Service?

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Shared/_AdditionalFilters.cshtml
```diff
@@ -75,42 +74,19 @@
...
-                            @if (@Model.InteroperabilityOptions[j].Text == @Model.InteroperabilityOptions[1].Text)
-                            {
-                                @for (var i = 0; i < Model.GPConnectIntegrationsOptions.Count; i++)
-                                {
-                                    <nhs-checkbox asp-for="@Model.GPConnectIntegrationsOptions[i].Selected"
-                                                  label-text="@Model.GPConnectIntegrationsOptions[i].Text"
-                                                  hidden-input="@Model.GPConnectIntegrationsOptions[i].Value" />
-                                }
-                            }
-                            @if (@Model.InteroperabilityOptions[j].Text == @Model.InteroperabilityOptions[2].Text)
```
dssherlock, line 106: I _think_ this was the loop that caused an exception on returning to the page for a Validation Error (it may have been a similar one if it wasn't this exact one). Check IntegrationTypes for null before Count call (or at least ensure they have been repopulated before returning the View)

src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/CompetitionResultsPdf/_FilterDetails.cshtml
```diff
@@ -145,32 +144,17 @@ else
         <nhs-summary-list-row label-text="Interoperability">
             @if (Model.HasInteroperabilityIntegrationTypes())
             {
-                @for (int j = 0; j < Model.FilterDetails.InteropIntegrationTypes.Count(); j++)
+                foreach (var integration in Model.FilterDetails.Integrations)
                 {
```
dssherlock, line 148: Nice-now get any new integrations for free! Can close off that ticket we raised for PDFs!

src/NHSD.GPIT.BuyingCatalogue.WebApp/wwwroot/js/GroupCounters.min.js
```diff
@@ -1 +1 @@
-function checkBoxClicked(n){updateGroupCount(n.target.getAttribute("subgroup"));updateTotals()}function updateGroupCount(n){const t=document.querySelectorAll(`.nhsuk-checkboxes__input[subgroup=${n}]:checked`).length,i=t>0?` - ${t} selected`:"";document.querySelector(`.counter-class[subgroup=${n}]`).innerText=i}function updateTotals(){const n=document.querySelectorAll(".nhsuk-checkboxes__input:checked").length,t=document.querySelectorAll(".nhsuk-checkboxes__input").length;Array.from(document.querySelectorAll(".selection-counter")).forEach(i=>i.innerText=`${n} out of ${t} selected`)}window.onload=function(){updateTotals();Array.from(document.querySelectorAll(`.nhsuk-checkboxes__input:checked`)).map(n=>n.getAttribute("subgroup")).filter((n,t,i)=>i.indexOf(n)===t).forEach(n=>updateGroupCount(n));Array.from(document.querySelectorAll(".selection-counter")).forEach(n=>n.removeAttribute("style"));Array.from(document.querySelectorAll(".nhsuk-checkboxes__input")).forEach(n=>n.addEventListener("change",checkBoxClicked,!1))};
```
dssherlock, line 1: Was this a deliberate change or one that got checked in by mistake? I feel this is one that often creeps in

tests/NHSD.GPIT.BuyingCatalogue.Framework.UnitTests/Extensions/SolutionExtensionsTests.cs
```diff
@@ -31,15 +31,6 @@ public static void GetFeatures_ReturnsDefaultFeaturesWhenNotSet(string features)
             result.Should().BeEquivalentTo();
         }
```
dssherlock, line 33: Again why?

tests/NHSD.GPIT.BuyingCatalogue.Services.UnitTests/Solutions/InteroperabilityServiceTests.cs
```diff
@@ -59,41 +57,60 @@ public static Task AddIntegration_NullIntegration_ThrowsException(Interoperabili
...
-           Integration newIntegration,
-           InteroperabilityService service)
+            Integration integration,
+            IntegrationType integrationType,
+            Solution solution,
+            [Frozen] BuyingCatalogueDbContext context,
+            InteroperabilityService service)
         {
-            solution.Integrations = JsonSerializer.Serialize(currentIntegrations);
-            solution.AdditionalServices.Clear();
```
dssherlock, line 69: Resharper really doesn't like this notation. The project builds ok but line 66 and 69 are being highlighted as invalid syntax. Not sure why. Actually if I'm honest i guess my question really is I'm not sure why it wouldn't.

tests/NHSD.GPIT.BuyingCatalogue.Services.UnitTests/ServiceHelpers/SolutionsFilterHelperTests.cs
```diff
@@ -69,433 +69,191 @@ public static void ParseCapabilityAndEpics_Capability_With_Epics()
...
-        public static void ParseApplicationTypeIds_EmptyAndWhiteSpace_GeneratesResults()
+        public static void ParseIntegrationAndTypeIds_IntegrationId_Types_Null()
         {
-            var input = "0.1. .2..    ";
+            var input = new Dictionary<SupportedIntegrations, int[]> { { SupportedIntegrations.Im1, null } };
 
-            var result = SolutionsFilterHelper.ParseApplicationTypeIds(input);
+            var result = SolutionsFilterHelper.ParseIntegrationAndTypeIds(input.ToFilterString());
 
-            var expected = new List<ApplicationType> { ApplicationType.BrowserBased, ApplicationType.Desktop, ApplicationType.MobileTablet };
```
dssherlock, line 102: Again resharper does not like this notation []

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/InteroperabilityModels/AddEditGpConnectIntegrationModel.cs
```diff
@@ -16,33 +15,46 @@ public AddEditGpConnectIntegrationModel()
         {
```
dssherlock, line 15: Feels there could be some refactoring of these 2 (maybe 3) models. Certainly GPConnect and IM1 are almost identical. Nhs App is slightly different coz it allows multiple integrations to be selected at the same time in checkboxes, rather than a dropdown.

src/NHSD.GPIT.BuyingCatalogue.Services/Solutions/InteroperabilityService.cs
dssherlock: Got it now! Multiselect rather than single select dropdown => multiple integrations being added/removed at the same time!

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Controllers/InteroperabilityController.cs
```diff
@@ -72,7 +70,10 @@ public async Task<IActionResult> AddIm1Integration(CatalogueItemId solutionId)
             if (solution is null)
```
dssherlock, line 70: All the Add/Edit/Delete actions for IM1 and GP Connect are so similar now. May be worth pulling these common actions out into a base class so will simplify the process when adding any new Integrations (as well as keeping the code neater instead of massive copy/paste job). An argument could also be made for the NHS App to be updated to follow a similar pattern rather than looking entirely different to the other two but that is more of a design change.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/SolutionsController.cs
```diff
@@ -119,8 +93,6 @@ public sealed class SolutionsController : Controller
         [HttpPost]
         public IActionResult Index(AdditionalFiltersModel model)
         {
```
dssherlock, line 95: Maybe worth just sanity checking that bug fix still works-i.e. in the non-js version of this page a user can select just an integration type without having to manually select the integration as well.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Models/InteroperabilityModel.cs
ghost: The item shouldn't be an associated or additional service since only solutions can have integrations associated with them. Currently, this model is used only within the context of a Solution.

This class and much of the Public Browse section was done at a time (~3 years ago) when `Solution` and `CatalogueItem` were used almost interchangeably.

What I might do is add a tech debt ticket to revisit this so that `Solution` is passed in where it should be and `CatalogueItem` is used only where the receiver can support all types of `CatalogueItem`

src/NHSD.GPIT.BuyingCatalogue.WebApp/wwwroot/js/GroupCounters.min.js
ghost: This one is an automated action but I've given up on fighting them. In another PR, I'll evalulate whether `GroupCounters` is actually required and remove it if not. If it is then I'll add it to the gitignore so it's produced only at build time.

tests/NHSD.GPIT.BuyingCatalogue.Framework.UnitTests/Extensions/SolutionExtensionsTests.cs
ghost: Answered above on related point

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Controllers/SolutionsController.cs
ghost: This is resolved via the parent-child hierarchy of the new Integration/Integration Types approach since the model first pulls back parents that are selected.

If a child is selected but a parent isn't, the parent and child are subsequently ignored. The child will be unchecked once the page reloads.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Controllers/InteroperabilityController.cs
ghost: > An argument could also be made for the NHS App to be updated to follow a similar pattern rather than looking entirely different to the other two but that is more of a design change.

I'd considered this for simplicity, but the result would be to move a lot of the service layer logic (handling stale records, adding new) to the controller which wouldn't be the right place for it. Although NHS App is just another integration, it's fundamentally different in how it's handled compared to GP Connect and IM1 which justifies a different approach to how it's saved since you can add one or many at any given time.

Was planning to tackle a rewrite of this controller separately. Rather than base and derived classes, I was thinking a generic method to handle the commit action and then some base/derived classes for the models which have different validation applied based on their derived types.

tests/NHSD.GPIT.BuyingCatalogue.Services.UnitTests/Solutions/InteroperabilityServiceTests.cs
ghost: Might be worth updating Visual Studio, .NET and Resharper to see if that resolves it. `[]` is the new collection initializer syntax released in C# 12 https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#collection-expressions

It's used in a few other places around the codebase at present

tests/NHSD.GPIT.BuyingCatalogue.Services.UnitTests/ServiceHelpers/SolutionsFilterHelperTests.cs
ghost: Resolving as answered above regarding syntax

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Admin/Models/InteroperabilityModels/AddEditGpConnectIntegrationModel.cs
ghost: Answered in comment regarding refactoring the way that interop is handled on the admin side

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Solutions/Views/Shared/_AdditionalFilters.cshtml
ghost: Updated to initialize the `IntegrationOptions` to a default of an empty list.

dssherlock (approved):

PR #1450, opened 2024-07-16
Migrate from Moq to NSubstitute

bjss-colin:
This completes the migration from Moq to NSubstitute and removes any NuGet references to Moq.


tests/NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests/ActionFilters/TermsOfUseActionFilterTests.cs
```diff
@@ -66,20 +66,24 @@ public static void Constructors_VerifyGuardClauses()
             executingContext.HttpContext.User = new ClaimsPrincipal(
                 new ClaimsIdentity(claims, "mock"));
 
-            nextDelegate.Setup(d => d())
-                .ReturnsAsync(executedContext);
```
dssherlock, line 70: This pattern with the internal NextDelegate function call seems to be used all over the place with just variations on the executing context. Any chance it can be pulled out?

ghost: Split out to an extension method for reusability

tests/NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests/Areas/Admin/Validators/ServiceLevelAgreementValidators/EditSLAContactModelValidatorTests.cs
```diff
@@ -14,67 +12,83 @@ namespace NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests.Areas.Admin.Validators.Serv
...
         [Theory]
-        [CommonAutoData]
+        [MockAutoData]
         public static void Validate_EmptyContactInformation_SetsModelError(
+            ServiceLevelAgreements serviceLevelAgreement,
+            [Frozen] IServiceLevelAgreementsService service,
             EditSLAContactModel model,
             EditSLAContactModelValidator validator)
         {
             model.ContactInformation = string.Empty;
```
dssherlock, line 41: Why have these been added? Or i guess the question is more how did the test pass before without it?

ghost: Moq automatically creates stubbed responses, which is what was also leading to tests being flaky and which led to some tests passing due to false positives.

NSubstitute has an opt-in model to automatically generate stubbed responses. That's been left off to ensure dependencies are explicitly configured to reduce false positives (where the test passes when data-flow hasn't been configured) and flaky tests (where randomly generated data flips between two states during test runs)

dssherlock: Ah ok good that makes sense. Coz i had experienced something similar occasionally myself when migrating tests.

tests/NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests/ModelBinders/TimeInputModelBinderTests.cs
```diff
@@ -12,76 +11,62 @@ namespace NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests.ModelBinders
     public static class TimeInputModelBinderTests
     {
         [Theory]
-        [InlineData(null)]
```
dssherlock, line 15: Not sure why removing the null check but if deliberate then needs removing from function name

dssherlock (dismissed):

tests/NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests/ModelBinders/TimeInputModelBinderTests.cs
ghost: That's a good point - test name updated.

Null check was removed because `ModelName` throws an exception on the setter. This isn't something we need to test since that's unrelated to this model binder.

dssherlock (approved):

PR #1451, opened 2024-07-17
Set Same Site Cookie policy to strict

dssherlock:


bjss-colin (approved):
Tested on PreProd over the App Gateway - looks good

PR #1454, opened 2024-07-18
Database cleanup

bjss-colin:
1. Remove the old Interop schema
2. Cleans up some unused logins and roles that are causing alerts on live platform


database/NHSD.GPITBuyingCatalogue.Database/Catalogue/Stored Procedures/FilterCatalogueItems.sql
```diff
@@ -1,131 +0,0 @@
-﻿CREATE PROCEDURE [catalogue].[FilterCatalogueItems]
```
dssherlock, line 1: I'm sure we have used this SP before. Not directly but when we are asked for data. May be useful to keep?

dssherlock (approved):

database/NHSD.GPITBuyingCatalogue.Database/Catalogue/Stored Procedures/FilterCatalogueItems.sql
ghost: It was previously used in the filtering journey but that logic was migrated to code-level filtering. I don't really see a point in keeping it since it's not being used - we can always go back through git history to restore the file if needed

PR #1461, opened 2024-08-02
Update NHS and GOV UK frontend package versions

bjss-colin:
This updates NHS and GOV UK Frontend packages to latest versions which included breaking changes to the Header and Footer components, both of which have been updated.

BuildBundlerModifier doesn't appear to be supported anymore so this has been swapped out for Gulp in the Dockerfile.

There are also tweaks to the homepage design to be more in line with the patterns used on the NHS.UK website - removing some classes that already exist under the NHS design system for alternating grey/white sections. `Your account` has been changed to `My account` (when signed in) matching the NHS site.

A logout link has been added back into the nav bar to be more accessible.

Several responsiveness issues have also been addressed with the homepage where elements overlap.

Currently, in live, the Menu and Login buttons overlap and the cards take up a large portion of the screenspace on mobile devices.

image

There's also a grey vertical bar to the right of the cards on smaller devices such as mobile or tablets.
image

Following the updates, the menu and account link no longer overlap since the account/login link sits within the same container as the logo. The cards take up less screenspace and have been reduced in width. The grey bar is now gone from beside the page. The `More` link uses the NHS' new component design.
image

Relevant links:

https://github.com/nhsuk/nhsuk-frontend/releases
https://service-manual.nhs.uk/design-system/components/header
https://service-manual.nhs.uk/design-system/components/footer


dssherlock (approved):
All generally looks fine. A couple of general accessibility issues I've found:

- Search component
Contrast error (not sure why styling even present as visually hidden)
Orphaned form label (can potentially remove label entirely)
- Skipped heading levels on several pages e.g. Catalogue Solutions, Orders page, Competitions page, Shortlist page
- Pagination component
Redundant link (due to page and Next link being next to each other)
- Save filters page broken Aria reference for Filter description
- Service Recipient page in general (wow)! Tbf most of these are regarding the table/checkboxes as discussed
image

PR #1470, opened 2024-08-28
Quantity error message

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.WebApp/Views/Shared/QuantitySelection/SelectServiceRecipientQuantity.cshtml
```diff
@@ -61,45 +65,46 @@
...
                     <nhs-table data-test-id="additional-service-dashboard-table">
                         <nhs-table-column>Organisation </nhs-table-column>
                         <nhs-table-column>ODS code</nhs-table-column>
                         <nhs-table-column>@Model.QuantityColumnTitle</nhs-table-column>
 
-                        @foreach (var recipient in location)
+                        @foreach (var recipient in Model.SubLocations[i].ServiceRecipients)
                         {
-                            var index = Array.FindIndex(Model.ServiceRecipients, x => x.OdsCode.Contains(recipient.OdsCode));
+                            var index = Array.FindIndex(Model.SubLocations[i].ServiceRecipients, x => x.OdsCode.Contains(recipient.OdsCode));
```
ghost: Let's use a nested for loop instead of `FindIndex`

src/NHSD.GPIT.BuyingCatalogue.WebApp/wwwroot/scss/_custom.scss
```diff
@@ -1,9 +1,13 @@
 .nhsuk-expander-index .nhsuk-details__summary .nhsuk-details__summary-text {
-  white-space: pre-line;
+    white-space: pre-line;
+}
+
+.nhsuk-details.nhsuk-expander.nhsuk-expander-index.nhsuk-error-summary {
```
ghost: I think the approach we talked about earlier was to create a new class that applies the red border. Let's do that instead if we can as it allows us to reuse that style elsewhere. This approach will require a new class for each component that we wish to add this style to.

I'm not too familiar with `unset` but, at face value, this potentially has an unexpected side effect of undoing all padding where some may be expected or wanted.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Validation/Quantity/ServiceRecipientQuantityModelValidator.cs
```diff
@@ -5,7 +5,7 @@ namespace NHSD.GPIT.BuyingCatalogue.WebApp.Validation.Quantity
 {
     public class ServiceRecipientQuantityModelValidator : AbstractValidator<ServiceRecipientQuantityModel>
     {
-        public const string ValueNotEnteredErrorMessage = "Enter a practice list size for {0}";
+        public const string ValueNotEnteredErrorMessage = "Enter a practice list size< for {0}>";
```
ghost: Validation can be tricky with our current implementation and we're definitely overdue a rework to make that system more modular. With that said, I'm wondering if using an attribute on the text input component might be an easier approach vs delimiting the error messages. Then it's only the text input tag builder that needs updated to append hidden text if it has been specified, and this error can become `Enter a practice list size`.

Part of the challenge with delimiting is that it can easily break if another developer were to inadvertently remove this as it's not immediately clear what it is/is doing.

bjss-colin (commented):
Haven't yet finished reviewing as I'm working on finishing my DPI changes also. Took a quick sneak peek

src/NHSD.GPIT.BuyingCatalogue.WebApp/wwwroot/scss/_custom.scss
dssherlock: It allows the css to use the inherited value (i.e. which is how the expander works) rather than the padding being explicitly set by the error class. But if we create a new error class for the border then we wont have that issue anyway.

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/TagHelpers/Helpers/TagHelperBuilders.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
+using System.Text.RegularExpressions;
```
ghost: Namespace is no longer in use following the refactoring

src/NHSD.GPIT.BuyingCatalogue.WebApp/Validation/Quantity/SubLocationModelValidator.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace NHSD.GPIT.BuyingCatalogue.WebApp.Validation.Quantity
+{
+    public class SubLocationModelValidator : AbstractValidator<SubLocationModel>
+    {
+        public const string ValueNotEnteredErrorMessage = "Enter all practice list sizes for {0}";
+
+        public SubLocationModelValidator()
+        {
+            RuleFor(x => x)
```
ghost, line 13: I think the validator is still giving up halfway through as there are too many errors with all of the individual recipients being tracked.

Neither of the 2 collapsed sublocations have quantities defined
image

image

The last errored sublocation has a point at which errors stop appearing against empty quantities
image

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.WebApp/Validation/Quantity/SubLocationModelValidator.cs
dssherlock: Have you tried this in Dev out of interest? Is this something new or did it do this before?

ghost: Tested locally just now.

It does give up part-way through in live, so it's an existing behaviour, but I'm wondering if there's something to be done that would at least highlight all sublocations that are missing quantities

bjss-colin (approved):

PR #1481, opened 2024-09-27
Feature/23690 review order

dssherlock:


src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/TagHelpers/DetailsAndExpander/ExpanderSummaryContentTagHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+    [HtmlTargetElement(TagHelperName, ParentTag = ExpanderSummaryTagHelper.TagHelperName)]
+    public sealed class ExpanderSummaryContentTagHelper : TagHelper
+    {
+        public const string TagHelperName = "nhs-expander-summary-content";
+
+        public const string TitleName = "title";
+        public const string TextName = "text";
+        public const string UrlTitleName = "url-title";
+        public const string UrlName = "url";
+        public const string JsOnlyName = "js-only";
```
ghost: It doesn't yet look like `JsOnly` is in use, as it's only used once we could potentially add the class directly in that view file. I wouldn't be too influenced by the current expander implementation as creating this new expander component should enable us to strip out lots of the bespoke logic that we had to add to the original Expander component due to how limiting it was.

With this new version, it should be dynamic to a degree that allows us to adapt and extend any individual page to meet the needs of that particular view, where the tag helper simply bootstraps the layout to a specification (the NHS Frontend). That way it prevents the need for a V3 in the future if it becomes too restrictive.

`Url` and `UrlTitle` may also be suitable candidates for handling in the view directly since that currently only happens on one page under Review - Planned Delivery Dates.

There should be short-hand attributes for setting the title in the 90% of scenarios where the title is straight forward. In the other 10% of scenarios we should be able to define HTML directly in the view to hand-craft that specific title, but if there comes a point where that style is used in more than one place then we can bake that into the tag builders in some way to streamline the process.

bjss-colin (changes requested):

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/TagHelpers/DetailsAndExpander/ExpanderSummaryContentTagHelper.cs
dssherlock: Yep i agree  completely

dssherlock: Thinking probably worth removing the ExpanderSummaryContentTagHelper completely and just putting the SecondaryText and SecondaryTitle properties back into the ExpanderSummaryTagHelper. If they want to use the default secondary content, then provide the secondary text and title, otherwise provide the html directly in the view. Thoughts?

dssherlock: Updated to remove ExpanderSummaryContentTagHelper as suggested. Added secondary content properties to ExpanderSummaryTagHelper. Can be rolled back if you think its better to include explicitly.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/Order/Summary.cshtml
```diff
@@ -184,16 +190,37 @@
...
-                }
+                        @if (Model.HasBespokeMilestones)
+                        {
+                            <div class="nhsuk-u-margin-bottom-9">
+                                <partial name="MilestoneTable" model="new MilestoneTableModel(Model.BespokeMilestoneLabelText, Model.BespokePlan.Milestones, false, Model.CallOffId, Model.InternalOrgId)" />
+                            </div>
+                        }
+                        else
+                        {
+                            <div class="nhsuk-u-margin-bottom-9">
```
ghost: Margin bottom results in excess whitespace. Let's remove this, from here and on line 202, to be consistent with other expanders on the page.

image

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/Order/Summary.cshtml
```diff
@@ -212,56 +239,78 @@
                         if (associatedServices.Count == 1)
                         {
                             totalDisplayed = true;
-                            itemModel.OrderTotals = Model.BuildOrderTotals();
+                            var orderTotals = Model.BuildOrderTotals();
+                            orderTotals.Inverted = true;
+                            itemModel.OrderTotals = orderTotals;
```
ghost: Let's remove the card as it does look a bit funky. The prices could be displayed directly in the expander

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/Order/Summary.cshtml
```diff
@@ -323,90 +372,92 @@
                             </nhs-table-cell>
                         </nhs-table-row-container>
                     </nhs-table>
-
+                    </nhs-card-content>
+                   </nhs-card-v2>
                 }
                 else
                 {
+                    <h2 id="costs-title">Cost of order</h2>
```
ghost: Should this be present for all types of orders? At the moment it's only visible for an original order

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/Order/Summary.cshtml
```diff
@@ -323,90 +372,92 @@
...
-        {
-            <div class="nhsuk-u-margin-top-3">
-                <vc:nhs-delete-button url="@Url.Action(nameof(OrderController.TerminateOrder), typeof(OrderController).ControllerName(), new { Model.InternalOrgId, Model.CallOffId })" text="Terminate this contract" />
-            </div>
-        }
 
         <div class="nhsuk-u-margin-top-3">
-            @if (Model.Order.OrderStatus == OrderStatus.InProgress
-            && Model.OrderWrapper.CanComplete())
+            @switch (Model.Order.OrderStatus)
```
ghost: As this page is becoming quite large with lots of different conditions, this could potentially be moved to the model which would allow for it to be unit tested

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/Order/_OrderTotal.cshtml
```diff
@@ -3,42 +3,44 @@
 @{
     var contractTerm = Model.MaximumTerm.HasValue ? $" ({Model.MaximumTerm} months)" : " (estimated)";
 }
-
-<h4 data-test-id="indicative-costs-title">Total indicative costs excluding VAT</h4>
-<nhs-summary-list>
-    <nhs-summary-list-row label-text="Total one-off cost:" data-test-id="one-off-cost-summary">
-        £@($"{Model.TotalOneOffCost:N2}")
-    </nhs-summary-list-row>
+<nhs-card-v2 inverted="@Model.Inverted">
```
ghost: Let's handle the card in the parent page as this partial should be agnostic of the implementation details. Adding a card here means every consumer of this partial is constrained to this specific design.

Handling this in the parent design also removes the need for an inverted card since that one scenario can display this data directly in the expander.

bjss-colin (changes requested):

bjss-colin (approved):

PR #1484, opened 2024-10-03
Migrate Org Importer to Function App

bjss-colin:
Migrates the Org Importer job to the function app which is set to trigger on a daily basis @ 8PM local to the server time.

Since the previous approach of using Entity Framework was quite slow, taking up to 30 minutes, this new approach implements `SqlBulkCopy` as the means of loading the data by first adding the data to a series of staging tables. Once the data is loaded then a stored procedure is executed which handles the merging of that data before finally truncating the staging tables.


src/BuyingCatalogueFunction/OrganisationImport/Services/TrudApiService.cs
```diff
@@ -0,0 +1,30 @@
...
+    IOptions<TrudApiOptions> trudApiOptions)
+    : ITrudApiService
+{
+    private readonly HttpClient httpClientFactory = httpClientFactory.CreateClient(nameof(TrudApiService));
+
+    private readonly TrudApiOptions trudApiOptions =
+        trudApiOptions?.Value ?? throw new ArgumentNullException(nameof(trudApiOptions));
+
+    public async Task<TrudApiResponse.Release> GetLatestReleaseInfo()
+    {
```
dssherlock, line 23: This feels a bit like a magic string-maybe this should be a setting in case it changes?

dssherlock (approved):

src/BuyingCatalogueFunction/OrganisationImport/Services/TrudApiService.cs
ghost: This is a URL path and shouldn't change unless there's a breaking change in the we're consuming API. In that event, we'd have to redeploy the app anyway.

PR #1491, opened 2024-11-11
Add "Expired" OrderStatus

ashleywntrbjss:
This PR:

- Adds a new Expired status to the order status enum

- Uses the existing Order.ContractExpired property to return the Expired value for Order.OrderStatus

- Status tag
    - adds a new 'orange' expired status and,
    - simplifies the duplicated logic for evaluating the tag colour / title

- Indirectly disables the terminate contract button for an expired order by no longer using the Complete status which the terminate button was tied to.

To be thorough, I also evaluated all references to OrderStatus to check for other logic that relied on this value, as a new status represents quite a substantial change.
Changes unrelated to the feature request but required due to refactor are as follows:

- Uses InProgress status for CanComplete, rather than !Complete status, as expired, terminated or deleted orders should also not be completed
    - Removes checks for InProgress status when used in conjunction with CanComplete 

- The linq queries for orders (GetOrders and GetOrdersBySearchTerm) now consider the expired status in the Take Until query as there shouldn't be any revisions after an expiration, as the CanBeAmended check already explicitly checks for expiration 

- Updates all instances of PDF naming logic to use an order status enum extension. Three existing implementations did it each slightly differently.

- Updates PDF cache lookup to skip the cache for in progress orders, but check it for all others - simplifies existing logic and factors in new expired status.

- Updates the test data to ensure a fresh set of test data doesn't cause an exception when loading orders (related to Order type test data not being present)


src/NHSD.GPIT.BuyingCatalogue.Framework/NHSD.GPIT.BuyingCatalogue.Framework.csproj
```diff
@@ -18,6 +18,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="8.0.1" />
```
dssherlock: Keep this change in its own feature branch

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/NHSD.GPIT.BuyingCatalogue.ServiceContracts.csproj
```diff
@@ -19,6 +19,7 @@
     <PackageReference Include="Azure.Storage.Blobs" Version="12.14.1" />
     <PackageReference Include="Azure.Storage.Queues" Version="12.12.0" />
```
dssherlock: And this one!

src/NHSD.GPIT.BuyingCatalogue.Services/NHSD.GPIT.BuyingCatalogue.Services.csproj
```diff
@@ -31,6 +31,7 @@
     <PackageReference Include="LinqKit.Microsoft.EntityFrameworkCore" Version="8.1.5" />
```
dssherlock: And this!

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/NHSD.GPIT.BuyingCatalogue.EntityFramework.csproj
```diff
@@ -38,6 +38,7 @@
     </PackageReference>
     <PackageReference Include="Enums.NET" Version="4.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
```
dssherlock: I think this should actually only be needed in the Regression test project. It shouldn't need adding to others

database/NHSD.GPITBuyingCatalogue.Database/PostDeployment/OrderSeedData/InsertTestOrderSeedData.sql
```diff
@@ -68,7 +68,7 @@ BEGIN
     -------------------------------------------------------
 
     INSERT INTO ordering.Orders
```
dssherlock, line 70: I'm not sure why we need to add the OrderTypeId on here for this change?

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/NHSD.GPIT.BuyingCatalogue.EntityFramework.csproj
ashleywntrbjss: Ah nuget said it was in every project.. I'll be more discerning next time! Have removed all of the webdriver updates

database/NHSD.GPITBuyingCatalogue.Database/PostDeployment/OrderSeedData/InsertTestOrderSeedData.sql
ashleywntrbjss: I updated this to address an exception that happens because this ID isn't included in the test data currently. As I was working in this area the frequent crashes were a bit of a chore. Could probably move it to the regression test branch or can raise a bug for it

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -249,7 +249,7 @@ public async Task<List<Order>> GetOrders(int organisationId)
                     .Where(o => o.OrderingPartyId == organisationId)
                     .ToListAsync())
                 .GroupBy(x => x.OrderNumber)
```
dssherlock, line 251: I've talked this through with Seana and I think this is correct. An order can have multiple amendments (revisions). However these will all have the same Commencement date. As a result, orders with a particular OrderNumber will all Expire at the same time. The logic below ensures only the most recent amendment of an Expired order will be displayed on the dashboard.

src/NHSD.GPIT.BuyingCatalogue.Services/Pdf/CachedOrderPdfService.cs
```diff
@@ -28,17 +29,22 @@ public async Task<MemoryStream> CreateOrderSummaryPdf(Order order)
...
-        if (order.OrderStatus != OrderStatus.Completed && order.OrderStatus != OrderStatus.Terminated)
+        if (order.OrderStatus == OrderStatus.InProgress)
             return await orderPdfService.CreateOrderSummaryPdf(order);
 
-        var blobDocument = order.OrderStatus == OrderStatus.Completed ? $"{order.CallOffId.ToString()}.pdf" : $"{order.CallOffId.ToString()}-terminated.pdf";
-        var cachedPdf = await azureBlobStorageService.DownloadAsync(new(settings.OrderPdfContainerName, blobDocument));
+        var callOffId = order.CallOffId.ToString();
+
+        var orderStatus = order.OrderStatus.AsFormattedString();
+
```
dssherlock, line 38: Changing the filename will have implications for all existing orders. Caching the pdf improves efficiency, but it was also introduced to maintain data integrity at the point the order was made i.e they user should always see the PDF as it appeared at the time of order. This is important as it outlines the contract. By changing the filename, existing orders will not be found in the blob storage and so a new one will be created. This would include any changes that have been made to the PDF format in the interim.

dssherlock, line 38: Changing the filename will have implications for all existing orders. Caching the pdf improves efficiency, but it was also introduced to maintain data integrity at the point the order was made i.e they user should always see the PDF as it appeared at the time of order. This is important as it outlines the contract. By changing the filename, existing orders will not be found in the blob storage and so a new one will be created. This would include any changes that have been made to the PDF format in the interim.

src/NHSD.GPIT.BuyingCatalogue.Services/Pdf/CachedOrderPdfService.cs
```diff
@@ -28,17 +29,22 @@ public async Task<MemoryStream> CreateOrderSummaryPdf(Order order)
     {
         if (order == null) throw new ArgumentNullException(nameof(order));
```
dssherlock, line 31: This condition is to allow "throw-away" versions of the PDF to be opened from within the order journey. These aren't cached. I think it is fine for it only to be allowed for InProgress orders. However, it will have a knock-on effect of allowing Deleted orders and Expired orders to be cached. Not sure if this is correct. Worth checking.

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/TagHelpers/Tags/NhsTagsTagHelper.cs
```diff
@@ -159,6 +159,8 @@ private static (TagColour SelectedColourClass, string TagText) GetOrderStatus(Or
         {
             var selectedColourClass = orderStatus switch
             {
```
dssherlock, line 161: Not sure who decides colours! Possibly worth checking the design guide.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/OrderController.cs
```diff
@@ -188,12 +189,9 @@ public async Task<IActionResult> Download(string internalOrgId, CallOffId callOf
 
             var result = await pdfService.CreateOrderSummaryPdf(order);
```
dssherlock, line 191: Wider discussion wrt allowing the download of other Status'

dssherlock (dismissed):
Looks good! A couple of questions wrt wider functionality and how the site should behave in some edge case scenarios-e.g. if an order is Deleted. May be worth talking through with Colin when he is back. But looks good.

database/NHSD.GPITBuyingCatalogue.Database/PostDeployment/OrderSeedData/InsertTestOrderSeedData.sql
dssherlock: Nope that's fine-I just couldn't work out how it related to the changes!

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/TagHelpers/Tags/NhsTagsTagHelper.cs
ashleywntrbjss: Thanks, do you know where I can find it?

ashleywntrbjss: Using red for expired as discussed

src/NHSD.GPIT.BuyingCatalogue.Services/Pdf/CachedOrderPdfService.cs
ashleywntrbjss: Reinstated previous file naming scheme to preserve cache

src/NHSD.GPIT.BuyingCatalogue.Services/Pdf/CachedOrderPdfService.cs
ghost: I think this is fine. Deleted orders aren't visible to the buyers and we don't expose PDF downloading capabilities to  admins for a deleted order.

In terms of an expired order, since the order will have been completed previously then there should be a pre-existing PDF in the cache from when it was originally completed. This logic means that an expired order will download the original order PDF which is fine since that's the buyers audit trail.

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
ghost: Another point on this. The reason the logic is this way is because an order is comprised of order revisions. Each amendment doesn't create a new order, instead they create revisions on an existing order.

This logic ensures that a user only sees the most recent completed, expired or terminated order revision - whether it be an original order revision or an amendment.

bjss-colin (approved):
Happy with this approach - nicely done

bjss-colin (approved):

PR #1493, opened 2024-11-19
Update Regression Test Suite

ashleywntrbjss:
This PR:

Updates the selenium chromedriver to work with the latest version of chrome

Updates the regression test suite to account for component changes

Updates the page loaded common action to tell you which page it was expecting rather than just ' false '

Splits some tests out into separate suites to allow running in parallel (though there seems to be a login issue - the first test will fail, all others will pass, and the first test will pass on re run)

Updates the test data to ensure that the single solution filter actually returns a single solution


tests/NHSD.GPIT.BuyingCatalogue.RegressionTests/Pages/Competitions/CompetitionPages.cs
```diff
@@ -126,6 +126,7 @@ public void StepOnePrepareCompetition(FilterType filterType, string competitionN
                 StartCompetition.CreateCompetition(competitionName);
                 SelectSolutions.AddSolutions(solutions);
                 SolutionShortlisted.SolutionNotIncludedInShortlisting();
```
dssherlock: Revert this file

tests/NHSD.GPIT.BuyingCatalogue.E2ETests.Framework/Actions/Common/CommonActions.cs
```diff
@@ -382,13 +384,20 @@ public bool PageLoadedCorrectGetIndex(
             var actionUrl = new Uri("https://www.fake.com/" + absoluteRoute, UriKind.Absolute);
```
dssherlock, line 385: Isn't this the check to determine if the test fails or passes? Hence it should return false, leading to the test failing gracefully. It should only throw an exception when something unexpected has happened.

tests/NHSD.GPIT.BuyingCatalogue.RegressionTests/LocalWebApplicationFactory.cs
```diff
@@ -33,7 +33,7 @@ public sealed class LocalWebApplicationFactory : WebApplicationFactory<Startup>
 
         // Need to find a better way of doing this
         [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1310:Field names should not contain underscore", Justification = "This name is used by the Webapp, so needs to be kept")]
```
dssherlock, line 36: This isn't necessary

tests/NHSD.GPIT.BuyingCatalogue.E2ETests.Framework/Actions/Common/CommonActions.cs
ashleywntrbjss: This is a bit of a makeshift way to get extra information out of this check without having to refactor every instance where its used. At the moment it just returns false and doesn't really tell you what went wrong. With the more detailed output I was able to tell that it was still in the process of logging in when it failed

tests/NHSD.GPIT.BuyingCatalogue.RegressionTests/LocalWebApplicationFactory.cs
ashleywntrbjss: For me localhost didn't work - I tried to keep it separate and uncommitted but I kept accidentally including it in things so I was hoping I'd be able to make the change here, then as long as it works for everyone else that should be the problem solved

tests/NHSD.GPIT.BuyingCatalogue.E2ETests.Framework/Actions/Common/CommonActions.cs
dssherlock: Does it not show up in the stack trace?

dssherlock (dismissed):

bjss-colin (approved):

PR #1495, opened 2024-12-03
Bugfix: Alignment for Associated Services across Order and Competition journeys

bjss-colin:
Resolves a bug whereby adding Associated Services to a competition would remove the previously added services. This was introduced during the order journey redesign towards the start of this year due to both journeys reusing the same shared views for adding services.

Since the order journey business logic changed, drifting from the competition journey logic, this meant that existing services were being hidden from the `Add Associated Services` page.

This change aligns both journeys again to make them consistent by porting the order logic over to the competition journey:
1. Remove existing services from the Add Services pages
2. Delete the `Add Service` confirmation page
3. Introduce a Delete link under each Service
4. Add the Delete confirmation page from the ordering journey


ashleywntrbjss (approved):
Provided offline feedback on remaining differences between associated services in the competition flow vs in the order flow (some whitespace and phrasing). Those were addressed, happy to approve.

PR #1497, opened 2024-12-05
Terraform: Add settings for secondary VPN

bjss-colin:
Introduces a new list variable for the secondary VPN IPs which is then mapped to the App Service and Deployment Slot (think Blue/Green) to permit access to the secondary VPN IP addresses via the [ip_restriction](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#action-2) block.

Uses a [dynamic block](https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks) to dynamically add an `ip_restriction` for each IP address. Starts the priority at 300 to create a buffer between static and dynamic entries.

### DIFF
--- a/terraform/webapp/webapp.tf
+++ b/terraform/webapp/webapp.tf
@@ -14,6 +14,7 @@ module "webapp" {
   aspnet_environment              = var.environment
   instrumentation_key             = azurerm_application_insights.appinsights.instrumentation_key
   primary_vpn                     = var.primary_vpn
+  secondary_vpn                     = var.secondary_vpn
   app_gateway_ip                  = module.appgateway.appgateway_pip_ipaddress
   app_dns_url                     = var.app_url
   docker_registry_server_url      = data.azurerm_container_registry.acr.login_server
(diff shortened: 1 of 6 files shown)


terraform/webapp/webapp.tf
```diff
@@ -14,6 +14,7 @@ module "webapp" {
   aspnet_environment              = var.environment
   instrumentation_key             = azurerm_application_insights.appinsights.instrumentation_key
   primary_vpn                     = var.primary_vpn
+  secondary_vpn                     = var.secondary_vpn
```
ashleywntrbjss: Extra space

ashleywntrbjss (dismissed):
Viewed outcome of terraform plan in platform dev release stage, confirmed adding IPs correctly

terraform/webapp/webapp.tf
ghost: Removed

ashleywntrbjss (approved):

PR #1501, opened 2024-12-11
Terraform: Upgrade versions and resolve more policy compliance

bjss-colin:
1. Updates Terraform and Azure RM versions
2. Adds the newer `immutable_backups_enabled` key for Azure SQL, defaulted to false
3. Updates the `probe` on the App Gateway instance to add the default values
4. On the function app, disable public access and default auth for SCM & FTP

The function app isn't accessed via the App Service, or any other service, so there shouldn't be a reason to preserve public access.


ashleywntrbjss (dismissed):

bjss-colin:
Planning on picking this up again in the new year as the storage account changes require `terraform import` to prevent the resource from being recreated

ashleywntrbjss (approved):

PR #1513, opened 2025-01-23
Add function to rebuild SQL indexes

bjss-colin:
Several indexes in the database have become fragmented due to bulk writes which is having an impact on performance for the website.

In the past, with on-premise, this sort of thing would've been added as an agent job. Microsoft have left a bit of a gap with Azure SQL since Agent jobs don't exist (they do with Managed SQL instances). There are Elastic Jobs but they only run on Azure SQL SKUs of S1 and above which prevents us from testing this in lower environments.

This pull request does the following:
1. Adds a stored procedure that'll rebuild any heavily fragmented indexes, or reorganize indexes that aren't as heavily fragmented.
2. Update the [Microsoft.Build.Sql](https://github.com/microsoft/DacFx/tree/main/src/Microsoft.Build.Sql) SDK to the latest version
3. Update the local docker image to pull the latest version of SqlPackage
4. Update the bash script to pass in two parameters for the newer version of SqlPackage
5. Add a Function that runs every Saturday @ 8PM to execute the stored procedure.

While making changes, I also took the libertry to:
1. Remove Serilog as this is producing double-log entries in App Insights
2. Tweak the default log levels as it's currently too verbose

Additionally, I refactored how service dependencies are registered in the IoC container. The idea is that all individual functions are self-contained units which can be added/removed without the need to refactor additional classes (aside from tests).

### DIFF
--- a/src/BuyingCatalogueFunction/DatabaseMaintenance/Services/DatabaseIndexMaintenanceService.cs
+++ b/src/BuyingCatalogueFunction/DatabaseMaintenance/Services/DatabaseIndexMaintenanceService.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using BuyingCatalogueFunction.DatabaseMaintenance.Interfaces;
 using Microsoft.EntityFrameworkCore;
@@ -7,6 +8,7 @@ using NHSD.GPIT.BuyingCatalogue.EntityFramework;
 
 namespace BuyingCatalogueFunction.DatabaseMaintenance.Services;
 
+[ExcludeFromCodeCoverage(Justification = "Executes stored procedure")]
 public class DatabaseIndexMaintenanceService(
     BuyingCatalogueDbContext dbContext,
     ILogger<DatabaseIndexMaintenanceService> logger) : IDatabaseIndexMaintenanceService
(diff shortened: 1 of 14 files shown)


src/BuyingCatalogueFunction/DatabaseMaintenance/Services/DatabaseIndexMaintenanceService.cs
```diff
@@ -0,0 +1,32 @@
...
+    ILogger<DatabaseIndexMaintenanceService> logger) : IDatabaseIndexMaintenanceService
+{
+    private readonly BuyingCatalogueDbContext dbContext = dbContext ?? throw new ArgumentNullException(nameof(dbContext));
+    private readonly ILogger<DatabaseIndexMaintenanceService> logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+    public async Task RebuildIndexes()
+    {
+        try
+        {
+            dbContext.Database.SetCommandTimeout(TimeSpan.FromMinutes(20));
```
ashleywntrbjss, line 23: Will this timeout take effect for the other services also accessing the dbcontext?

ghost: In theory it shouldn't because the context and context options are registered as scoped services rather than singleton instances. The timeout should reset with each function invocation

ghost: Tested this and can see that the timeout is resetting to the default value of `null` which means it'll use the underlying connection timeout value.

ashleywntrbjss (approved):

PR #1519, opened 2025-01-29
Bug: use correct Integration ID for delete URL

bjss-colin:
Resolves 2 bugs:
1. The incorrect ID is being used as part of the Delete Integration URL on the Admin section
2. The integration types aren't re-populated when POST results in an invalid model state


ashleywntrbjss (approved):
All unit and regression tests pass and bugfix works as expected

PR #1522, opened 2025-02-12
Bugfix: Allow users to visit Non-price elements weightings page once criteria has been reviewed

bjss-colin:
Removes the `CriteriaReviewedGuardFilter` attribute from the `Weights` action on `CompetitionNonPriceElementsController` as this filter should be used only when there isn't a read-only view of a page where the user would be able to change values up to that point


ashleywntrbjss (approved):
Tested working

PR #1523, opened 2025-02-12
Bugfix: Add validation for tiered price lower range

bjss-colin:
Adds a validation scenario for when the lower range is <= 0.

The error message used is similar to existing greater than 0 errors


ashleywntrbjss (approved):
Tested working

PR #1527, opened 2025-02-25
Task List accessibility improvements

bjss-colin:
This pull request updates the NHSUK Frontend version reference and swaps the bespoke task list items out for the newer NHS task list item design which brings several accessibility improvements.

The task list items are now clickable elements when there's a link attached. Previously it was only the text that was clickable.

Screen readers will now read out each task list item in order of
1. Label text
2. Label hint
3. Status

As per the [task list design](https://service-manual.nhs.uk/design-system/components/task-list), screen readers do not read items that cannot yet be started.

This drifts slightly from the NHS Design in that it retains the status tags for all status types, whereas the NHS design seems to only display status tags when something is in-progress.

Version 9 of the NHSUK Frontend components brings a suggested change for Backlinks which would be a massive change from what we currently do. It suggests moving the backlink to outside the `<main></main>` tags. We should look at this in the future if it becomes an issue.

https://github.com/nhsuk/nhsuk-frontend/releases/tag/v9.0.0


ashleywntrbjss:
AC says this about screen reader 

_Given the user tabs to Status
Then the screen reader should read e.g. "Status In Progress"_

At the moment this just reads the status directly, not sure if we need 'status' prefix as well.

ashleywntrbjss:
Could add this to ignore

image

ashleywntrbjss (approved):
Test evidence in ticket and comments addressed

PR #1529, opened 2025-03-10
Add Created column to users table

bjss-colin:


ashleywntrbjss (approved):
Created matches expected time when not pre populated (start time of the earliest entity). Created set as expected when creating new 
image

PR #1534, opened 2025-04-30
Move framework expiry to standalone page

bjss-colin:
* Moves the framework expiry callout to a standalone page
* Removes the `Disabled` support for `SelectOption<T>`
* Removes expired frameworks for the order triage journey
* Prevents order creation if there are no procurement frameworks


ashleywntrbjss (approved):
Approved - see test evidence in tickets

PR #1537, opened 2025-05-22
Improve performance of amend journey

bjss-colin:
The amend journey has an enormous memory footprint due to the deep-cloning implementation whereby orders are serialized as JSON and deserialized in order to clone the top level order and all complex objects beneath it. This happens multiple times per page resulting in a whopping 5GB memory footprint

image

This resulted in very long response times
image

### All orders were initially being tracked by the Entity Framework change tracker
To solve this, I've split the query behind the `OrderWrapper` out to two separate queries.
1. The previous orders are explicitly queried with tracking disabled
2. To maintain backwards compatibility, the Order object is tracked

### The cloning approach was a very inefficient approach relying on serialization/deserialization
This has been replaced with an `ICloneable<T>` interface for explicit implementations. I've avoided the default .NET `ICloneable` as it would require boxing/unboxing for value types since it uses `object`.

After these two changes, the memory footprint dropped significantly

image

with much more reasonable response times
image

### The full order trail is being pulled in scenarios where it isn't needed
There are many places where `OrderWrapper` is being populated with the current order and all previous orders when only a single property value from the current order is used, such as `DeliveryDate`. This is a massive drain on resources.

I haven't explicitly tackled this issue in this pull request but, after some brief inspections of the codebase, I believe there's a path where we
1. Understand the current architecture of the amend journey
2. Implement new service methods to query order data directly
3. Work towards incorporating the amend/order wrapper logic in SQL via projections
4. Use the order wrapper/"flattened" concept only on order summary pages

Additionally, I've removed some backwards compatibility methods (`EnsureOrderItemsForAmendment`) primarily because all orders will be in an expired state by time the new frameworks launch so these methods shouldn't be required.


ashleywntrbjss (approved):
Tested locally before and after, performance gains were very evident (proof in ticket #24101)

PR #1538, opened 2025-05-30
Allow standards compliance to be adjusted without work-off plans

bjss-colin:
This adds support for Standards to be marked as In Progress without having an explicit work-off plan associated with them.

Given the performance issues we've been facing in the order journey, namely due to several service methods being used in many scenarios, I've taken this as an opportunity to implement some more projection based queries.

Given there are only two statuses that represent the compliance of a Standard on a Solution (those statuses being In Progress and Fully Met), Solutions have a new associated table called `InProgressStandards`. The presence of a standard in this table marks the Standard compliance as being in progress.

The Standard selector on the Work-off Plans admin page is now pre-populated with those standards present in the `InProgressStandards` table. A standard must first be marked as In Progress before a work-off plan can be added.

Work-off plans for a standard will be removed if the standard status is changed from In Progress to Fully Met

The marketing pages will display the standards status based on this compliance model and a link to view the work-off plans will appear if there are work-off plans added.

There's a post-deploy migration script which adds relevant standards to the `InProgressStandards` table for every solution that has a work-off plan


ashleywntrbjss (dismissed):
Approved, see ticket for testing notes

ashleywntrbjss (approved):

PR #1541, opened 2025-06-11
24065 - Sublocations for Orders - into feature branch

ashleywntrbjss:
This PR:

Merges the order sublocation work into the sublocation feature branch, and makes some improvements to the competition implementation.

The competitions changes are as follows:

- Moves annotations to ef builder properties for consistency 
- Uses linq based hash set logic to avoid having to instantiate extra sets
- Uses variables to reduce duplication in migration script

The order sublocation work is as follows:

- Introduces new schema design for orders to introduce a sublocation layer in the order flow.
- Migrates order recipients into new tables, including joins to create the sublocation structure using the existing relationship data.
- Updates other services such as solution selection and quantity to work with the flattened recipients enumerable made available from the order entity. Keeping the flat structure reduces the amount of work required in this initial step.
- Includes new screens for edge cases, mainly around the amendment journey.


src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Ordering/Models/Order.Custom.cs
```diff
@@ -277,43 +301,61 @@ public void InitialiseOrderItemsFrom(ICollection<OrderItem> items)
...
+            Order previous,
+            CatalogueItemId catalogueItemId)
+        {
+            return cr =>
+            {
+                OrderSublocationRecipient previousRecipient =
+                    previous.FlattenedRecipients.FirstOrDefault(pr =>
+                        pr.RecipientOdsCode == cr.RecipientOdsCode);
+
+                var previousRecipientDidNotExist = previousRecipient is null;
```
ghost: The subsequent lines could be turned into a one-liner

```csharp
return previousRecipient is null || previousRecipient.OrderItemSublocationRecipients.All(oir => oir.CatalogueItemId != catalogueItemId);
```

src/NHSD.GPIT.BuyingCatalogue.Services/Contracts/DeliveryDateService.cs
```diff
@@ -65,17 +65,18 @@ public async Task ResetRecipientDeliveryDates(int orderId)
 
         public async Task SetDeliveryDates(int orderId, CatalogueItemId catalogueItemId, List<RecipientDeliveryDateDto> deliveryDates)
         {
-            var recipients = await dbContext.OrderRecipients.Include(x => x.OrderItemRecipients)
-                .Where(x => x.OrderId == orderId)
+            List<OrderItemSublocationRecipient> orderItemSublocationRecipients = await dbContext
```
ghost, line 68: Nice one - this is a cleaner approach and pulls back less data

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderQuantityService.cs
```diff
@@ -62,42 +63,46 @@ public async Task SetOrderItemQuantity(int orderId, CatalogueItemId catalogueIte
...
-                    recipient.SetQuantityForItem(catalogueItemId, quantity.Quantity);
-                }
+                recipient.SetQuantityForItem(catalogueItemId, quantity.Quantity);
             }
 
             await dbContext.SaveChangesAsync();
         }
 
-        public async Task SetServiceRecipientQuantitiesToSameValue(int orderId, CatalogueItemId catalogueItemId, int quantity)
+        public async Task SetServiceRecipientsSingleQuantity(int orderId, CatalogueItemId catalogueItemId, int quantity)
```
ghost: This method could potentially be an overload of `SetServiceRecipientQuantities` since both do the same thing (setting quantities) and have different signatures

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -288,7 +572,7 @@ public async Task<List<Order>> GetOrders(int organisationId)
                     .Where(o => o.OrderingPartyId == organisationId)
                     .ToListAsync())
                 .GroupBy(x => x.OrderNumber)
-                .SelectMany(x => x.OrderByDescending(y => y.Revision).TakeUntil(y => y.OrderStatus is OrderStatus.Completed or OrderStatus.Terminated or OrderStatus.Expired))
+                .SelectMany(x => MoreEnumerable.TakeUntil(x.OrderByDescending(y => y.Revision), y => y.OrderStatus is OrderStatus.Completed or OrderStatus.Terminated or OrderStatus.Expired))
```
ghost: Out of interest, what does this do differently to the original?

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -328,7 +612,7 @@ public async Task<IList<SearchFilterModel>> GetOrdersBySearchTerm(int organisati
...
```
ghost: Out of interest, what does this do differently to the original?

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -280,6 +301,269 @@ public async Task<OrderWrapper> GetOrderForTaskListStatuses(CallOffId callOffId,
...
+                .Include(o => o.OrderingPartyContact)
+                .Include(o => o.Supplier)
+                .Include(o => o.LastUpdatedByUser)
+                .Include(o => o.OrderItems)
+                .ThenInclude(i => i.CatalogueItem)
+                .Include(o => o.SelectedFramework)
+                .Include(o => o.OrderSublocations)
+                .ThenInclude(os => os.SublocationOrganisation)
+                .AsSplitQuery()
+                .AsNoTracking()
```
ghost, line 313: `AsNoTracking` specified here and on `previousOrders`. Line #318 should be sufficient since this will apply to both queries

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -280,6 +301,269 @@ public async Task<OrderWrapper> GetOrderForTaskListStatuses(CallOffId callOffId,
             return new OrderWrapper(order, previousOrders);
         }
 
+        public async Task<OrderWrapper> GetOrderWithSublocations(CallOffId callOffId, string internalOrgId)
+        {
+            IQueryable<Order> orders = dbContext.Orders
+                .Include(o => o.AssociatedServicesOnlyDetails.Solution)
```
ghost: Might be worth sanity checking if all of the includes are required

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -280,6 +301,269 @@ public async Task<OrderWrapper> GetOrderForTaskListStatuses(CallOffId callOffId,
...
+        public async Task<bool> GetOrderHasAnySublocations(CallOffId callOffId, string internalOrgId)
+        {
+            return await dbContext.Orders
+                .Where(o => o.OrderNumber == callOffId.OrderNumber
+                    && o.Revision <= callOffId.Revision
+                    && o.OrderingParty.InternalIdentifier == internalOrgId)
+                .AnyAsync(o => o.OrderSublocations.Count > 0);
+        }
+
+        public async Task SetSublocations(
```
ghost, line 375: Let's do some performance profiling on the queries produced by this - checking both application memory and the execution plans in SSMS

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -280,6 +301,269 @@ public async Task<OrderWrapper> GetOrderForTaskListStatuses(CallOffId callOffId,
...
+                throw new InvalidOperationException(
+                    "One or more requested Ids not found or not valid for this organisation.");
+            }
+
+            HashSet<string> orderSublocations =
+                order.OrderSublocations.Select(x => x.SublocationOdsCode).ToHashSet();
+
+            HashSet<string> removes = orderSublocations.Except(sublocationOdsCodes).ToHashSet();
+
+            if (order.IsAmendment)
```
ghost: Is it possible to hit this scenario?

There's quite a lot happening in this method so I'd be keen to see where we can reduce potentially unnecessary database hits

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderSublocationService.cs
```diff
@@ -0,0 +1,170 @@
...
+                    "One or more requested Ids not found or not valid for this sublocation.");
+            }
+
+            OrderWrapper wrapper = await orderService.GetOrderWithCatalogueItemAndPrices(
+                sublocation.Order.CallOffId,
+                sublocation.Order.OrderingParty.InternalIdentifier);
+
+            HashSet<string> removes = currentRecipientsOdsCodes.Except(newRecipientOdsCodes).ToHashSet();
+
+            if (wrapper.Order.IsAmendment)
```
ghost: Need to determine whether it's possible to hit this scenario through the normal user flow. Validation is good but sometimes too much validation can result in bigger performance hits

src/NHSD.GPIT.BuyingCatalogue.Services/Organisations/TrudOdsService.cs
```diff
@@ -169,6 +170,19 @@ public async Task<IReadOnlyList<ServiceRecipient>> GetServiceRecipientsByParentI
         return serviceRecipients;
     }
 
+    public async Task<string> GetOrganisationName(string odsCode)
+    {
+        if (string.IsNullOrEmpty(odsCode))
```
ghost: `ArgumentException.ThrowIfNullOrWhiteSpace` would work here quite nicely

src/NHSD.GPIT.BuyingCatalogue.Services/TaskList/Providers/DeliveryDatesStatusProvider.cs
```diff
@@ -17,10 +18,26 @@ public TaskProgress Get(OrderWrapper wrapper, OrderProgress state)
...
+                return TaskProgress.CannotStart;
+            }
+
+            var anyDeliveryDatesEntered = order.OrderItems
+                .Any(x =>
+                {
+                    ICollection<OrderSublocationRecipient> recipients = wrapper
+                        .DetermineOrderRecipients(x.CatalogueItemId);
+
+                    if (recipients.Count == 0)
```
ghost: This might be redundant as the `.Any` on Line 38 should return false if there are no items

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsBacklink/NhsBacklinkViewComponent.cs
```diff
@@ -8,6 +8,16 @@ public sealed class NhsBacklinkViewComponent : ViewComponent
     {
         public async Task<IViewComponentResult> InvokeAsync(string url, string text)
         {
+            if (string.IsNullOrEmpty(url))
```
ghost: `ArgumentException.ThrowIfNullOrWhiteSpace` might be cleaner here

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsBacklink/NhsBacklinkViewComponent.cs
```diff
@@ -8,6 +8,16 @@ public sealed class NhsBacklinkViewComponent : ViewComponent
...
+            {
+                throw new ArgumentException($"{nameof(url)} cannot be null or empty");
+            }
+
+            if (string.IsNullOrEmpty(text))
```
ghost: `ArgumentException.ThrowIfNullOrWhiteSpace` might be cleaner here

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
```diff
@@ -280,6 +301,269 @@ public async Task<OrderWrapper> GetOrderForTaskListStatuses(CallOffId callOffId,
...
+                .AsNoTracking()
+                .Where(o => o.Revision < callOffId.Revision)
+                .ToListAsync();
+
+            Order order = await orders
+                .FirstOrDefaultAsync(o => o.Revision == callOffId.Revision);
+
+            var wrapper = new OrderWrapper(order, previousOrders);
+
+            var orderIsEditable =
```
ghost: Validation is typically multi-layered which means that an `ActionFilter` might be used at the controller level to prevent access when certain conditions aren't met. To reduce the amount of validation and database hits in service methods, we should first see if any filters exist that would prevent scenarios from being hit.

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Orders/OrderWrapper.cs
```diff
@@ -68,17 +67,57 @@ public OrderWrapper(Order currentOrder, IEnumerable<Order> previousOrders)
...
+                {
+                    return Order.OrderItems.Count > 0;
+                }
 
-        public bool HasNewOrderItems => Order.OrderItems
-            .Any(r => Previous?.OrderItems?
-            .FirstOrDefault(x => x.CatalogueItemId == r.CatalogueItemId) == null);
+                if (Order?.OrderItems is null || !Order.OrderItems.Any())
+                {
+                    throw new InvalidOperationException("Order items is null or empty");
```
ghost: One thing we need to be careful of is changing too much logic unrelated to the core feature as it's the core feature that we're primarily focused on testing. In this particular instance we're introducing newer exception based flow over the older true/false.

Rather than exceptions for control flow, I think returning true/false in these scenarios would be fine. If OrderItems is empty then it's technically correct that there are no new order items.

On the other hand, if the Previous item is null or there are no Previous OrderItems, but the current order revision has OrderItems, then it would be correct to say that there are newer order items associated with that order.

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Orders/OrderWrapper.cs
```diff
@@ -68,17 +67,57 @@ public OrderWrapper(Order currentOrder, IEnumerable<Order> previousOrders)
...
+        public bool HasNewOrderRecipients
+        {
+            get
+            {
+                if (Order.Revision <= 1)
+                {
+                    return Order.FlattenedRecipients.Any();
+                }
+
+                if (Order?.FlattenedRecipients is null || !Order.FlattenedRecipients.Any())
```
ghost: Null conditional operator shouldn't be required here as `Order` is validated in the constructor. Note: remove the `set;` from the `Order` property for this to be safe at compile-time.

Additionally, same comment as below applies here with regards to the newer exception control flow

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/ImportServiceRecipientsController.cs
```diff
@@ -94,113 +94,306 @@ await importService.Store(
...
+        List<(string Expected, string Actual, string OdsCode)> mismatchedNames =
+            GetMismatchedNames(cachedRecipients.ToList(), organisationServiceRecipients);
+
+        var shouldShowValidateNamesScreen = mismatchedNames.Count > 0;
+
+        if (shouldShowValidateNamesScreen)
+        {
+            OrderWrapper wrapper = await orderService.GetOrderThin(callOffId, internalOrgId);
+
+            var orderDescription = wrapper.Order.Description;
```
ghost: The description can have up to 100 chars. In the past we settled with the Call-off ID as the length is deterministic

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/ImportServiceRecipientsController.cs
```diff
@@ -94,113 +94,306 @@ await importService.Store(
...
+
+            var model = new ValidateOdsModel(
+                cachedRecipients.Where(x => mismatchedOdsCodes.Contains(x.OdsCode)))
             {
-                BackLink = Url.Action(nameof(Index), new { internalOrgId, callOffId }),
-                Caption = callOffId.ToString(),
-                CancelLink = Url.Action(nameof(CancelImport), new { internalOrgId, callOffId }),
+                BackLink = backAndCancelLink,
+                Caption = orderDescription,
```
ghost, line 152: The description can have up to 100 chars. In the past we settled with the Call-off ID as the length is deterministic

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/ImportServiceRecipientsController.cs
```diff
@@ -94,113 +94,306 @@ await importService.Store(
...
+                OdsCode = x.Key,
+                ServiceRecipients = x.Select(y => new ServiceRecipientModel(y))
+                    .ToList(),
+            })
+            .ToList();
+
+        var cancelLink = Url.Action(nameof(CancelImport), new { internalOrgId, callOffId });
+
+        var model = new ValidationCompleteModel(
+            orderDescription,
```
ghost, line 232: The description can have up to 100 chars. In the past we settled with the Call-off ID as the length is deterministic

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderSublocationService.cs
```diff
@@ -0,0 +1,170 @@
...
+                    "Sublocations cannot be edited for this order.");
+            }
+
+            IEnumerable<ServiceRecipient> validRecipientsForSublocation =
+                await odsService.GetServiceRecipientsBySublocation(sublocationOdsCode);
+
+            HashSet<string> currentRecipientsOdsCodes =
+                sublocation.SublocationRecipients.Select(x => x.RecipientOdsCode).ToHashSet();
+
+            var allIdsValid = newRecipientOdsCodes.All(x => validRecipientsForSublocation.Any(y => y.OrgId == x));
```
ghost, line 104: The database will perform validation of IDs via primary/foreign key constraints so this may not be necessary

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderSublocationService.cs
```diff
@@ -0,0 +1,170 @@
...
+                .OrderSublocations
+                .Where(
+                    OrderSublocationPrimaryKeyPredicate(parentOdsCode, orderId, sublocationOdsCode))
+                .Include(x => x.Order)
+                .ThenInclude(y => y.OrderingParty)
+                .Include(x => x.SublocationOrganisation)
+                .Include(x => x.SublocationRecipients)
+                .FirstAsync();
+
+            var hasSubsequentRevisions = await orderService.HasSubsequentRevisions(sublocation.Order.CallOffId);
```
ghost, line 81: This is something we could achieve globally with an action filter

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/QuantityController.cs
```diff
@@ -131,25 +130,37 @@ public async Task<IActionResult> SelectServiceRecipientQuantity(
...
-                    organisationRecipients.Location));
-
-            IEnumerable<ServiceRecipientQuantityDto> previousRecipients = wrapper.Previous?.OrderRecipients?.Select(
-                x => new ServiceRecipientQuantityDto(
-                    x.OdsCode,
-                    x.OdsOrganisation?.Name,
+
+            List<ServiceRecipientQuantityDto> recipientDtos = [];
+
+            Dictionary<string, string> sublocationNameCache = [];
```
ghost: If the Parent Organisation is included in the service query, this could be reduced to:

```csharp
List<ServiceRecipientQuantityDto> recipientDtos = orderRecipients
                .GroupBy(x => x.ParentSublocation.SublocationOrganisation.Name)
                .SelectMany(x => x.Select(y =>
                    new ServiceRecipientQuantityDto(
                        y.RecipientOdsCode,
                        y.RecipientOdsOrganisation?.Name,
                        y.GetQuantityForItem(orderItem.CatalogueItemId),
                        x.Key)))
                .ToList();
```

It's potentially worth having `DetermineOrderRecipients` preserve the parent-child structure but that's one we could pick up in the future

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/Contracts/DeliveryDatesController.cs
```diff
@@ -172,14 +172,7 @@ public async Task<IActionResult> EditDates(string internalOrgId, CallOffId callO
                 orderWrapper,
                 new RouteValues(internalOrgId, callOffId, catalogueItemId) { Source = source });
 
-            var orderRecipients = orderWrapper.DetermineOrderRecipients(catalogueItemId);
```
ghost, line 175: I've observed an issue whereby no delivery dates are being populated for an additional service that is added during an amendment.

For example:
Revision 1 - Solution with recipients
Revision 2 - extra recipients added to Solution
Revision 3 - added Additional service

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/ConfirmSublocationRecipientsModel.cs
```diff
@@ -14,14 +19,89 @@ public ConfirmSublocationRecipientsModel(
...
+        {
+            Caption = wrapper.Order.Description;
+            Advice = OrderAdvice;
+
+            IEnumerable<OrderSublocation> sublocationsWithNewRecipients = wrapper.Order.OrderSublocations.Where(x =>
+            {
+                OrderSublocation previousSublocation = wrapper.Previous?.OrderSublocations
+                    .FirstOrDefault(y => y.SublocationOdsCode == x.SublocationOdsCode);
+
+                if (previousSublocation is null)
```
ghost: Could be shortened to a single return statement since the recipient count will be greater than the previous if the previous is null - defaulting to 0

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/RemoveSublocationsModel.cs
```diff
@@ -14,7 +15,31 @@ public RemoveSublocationsModel(
...
+        }
+
+        public RemoveSublocationsModel(
+            Order order,
+            IReadOnlyList<string> sublocations,
+            IReadOnlyList<string> removes,
+            string backLink)
+            : this(sublocations, removes, backLink)
+        {
+            Caption = order.Description;
```
ghost: The description could be large. Also, in some pages related to the recipients we're using the Call-off ID while in others we're using the description. Personal preference would be to standardise on the Call-off ID

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/NoNewRecipientsForAmendmentModel.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+        }
+
+        public NoNewRecipientsForAmendmentModel(
+            Order order,
+            string backLinkUrl,
+            string continueLinkUrl)
+        {
+            Title = "Confirm service recipients";
+            Caption = order.Description;
```
ghost: Let's use Call-off ID here for caption

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/SelectSublocationsModel.cs
```diff
@@ -13,31 +14,94 @@ public SelectSublocationsModel()
...
+
+        private static List<SelectOption<string>> GetRenderedSublocations(
+            IEnumerable<ServiceModels.OdsOrganisation> possibleSublocations,
+            ICollection<SublocationModel> existingSublocations)
+        {
+            return possibleSublocations
+                .Select(sl =>
+                {
+                    var selected = existingSublocations.Select(es => es.OdsCode).Contains(sl.OdsCode);
+                    return new SelectOption<string> { Text = sl.OdsCode, Value = sl.OdsCode, Selected = selected };
```
ghost: We should be able to use the `SelectOption(string text, TValue value, bool selected)` constructor here

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/SelectSublocationsModel.cs
```diff
@@ -13,31 +14,94 @@ public SelectSublocationsModel()
...
+        {
+            return possibleSublocations
+                .Select(sl =>
+                {
+                    var selected = previousOrderSublocations.Any(ps => ps.OdsCode == sl.OdsCode)
+                        || existingSublocations.Any(es => es.OdsCode == sl.OdsCode);
+
+                    var sublocationPartOfPreviousOrder = previousOrderSublocations.Any(x => x.OdsCode == sl.OdsCode);
+
+                    return new SelectOption<string>
```
ghost: Worth adding a new constructor for `hidden` and using that in place here

bjss-colin (changes requested):
Overall it looks good. Comments based on discoveries during review and testing.

In some places I'd say there might be a bit too much validation taking place which is resulting in ~100 line methods. In instances like that, it's worth employing various validation techniques rather than having all validation occur in a single method. For example, the database will validate that there are no key violations (ODS IDs)

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
ashleywntrbjss: This appears to have come from the introduction of ToHashSet() in other methods in the service. With the generic import of MoreLinq, ToHashSet is ambiguous between the default and morelinq offering so I'm assuming the IDE handled this by refactoring the MoreLinq usage to be more explicit.

ashleywntrbjss: Removed use of ToHashSet and restored previous takeuntil

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
ashleywntrbjss: Removed use of ToHashSet and restored previous takeuntil

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
ashleywntrbjss: Remove redundant asnotracking

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Ordering/Models/Order.Custom.cs
ashleywntrbjss: Implemented one liner

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderQuantityService.cs
ashleywntrbjss: Implemented overloaded signature

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
ashleywntrbjss: Stripped out unused includes

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
ashleywntrbjss: Remove validation for amendment, csv import performs its own validation, for manual selection - opens up small edge case where url parameters for remove sublocation could be changed but low risk.

src/NHSD.GPIT.BuyingCatalogue.Services/Organisations/TrudOdsService.cs
ashleywntrbjss: Implement concise exception

src/NHSD.GPIT.BuyingCatalogue.Services/TaskList/Providers/DeliveryDatesStatusProvider.cs
ashleywntrbjss: Remove redundant count check

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsBacklink/NhsBacklinkViewComponent.cs
ashleywntrbjss: Implemented concise exception

src/NHSD.GPIT.BuyingCatalogue.UI.Components/Views/Shared/Components/NhsBacklink/NhsBacklinkViewComponent.cs
ashleywntrbjss: Implemented concise exception

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/ImportServiceRecipientsController.cs
ashleywntrbjss: Use call off id for caption

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/ImportServiceRecipientsController.cs
ashleywntrbjss: Use call off id for caption

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/ImportServiceRecipientsController.cs
ashleywntrbjss: Use call off id for caption

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderSublocationService.cs
ashleywntrbjss: Determining whether a service recipient is valid for a sublocation requires checking the relationship table so not sure if that could be done by keys alone. 

Although, the user gets presented with already validated options when selecting manually and the csv service should also validate sublocations when de-flattening so could do without this but just wanted to clarify its purpose.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/RemoveSublocationsModel.cs
ashleywntrbjss: Replaced with call off id

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/SelectSublocationsModel.cs
ashleywntrbjss: Switched to constructor

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/SelectSublocationsModel.cs
ashleywntrbjss: Switched to constructor

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/ConfirmSublocationRecipientsModel.cs
ashleywntrbjss: Replaced with single return

src/NHSD.GPIT.BuyingCatalogue.WebApp/Models/Shared/ServiceRecipientModels/NoNewRecipientsForAmendmentModel.cs
ashleywntrbjss: Using call off id

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/QuantityController.cs
ashleywntrbjss: GetOrderWithOrderItems has a lot of uses and this is the only one that requires the parent organisation, so didn't want to increase the burden on all of the other requests, but could do some profiling to see if this has any benefit.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/Contracts/DeliveryDatesController.cs
ashleywntrbjss: Initial investigation shows quantity is saving correctly in the db but the status isn't being accurately shown in the solutions and services section. On the main task list, solutions and services correctly shows as amended so you can progress to planned delivery dates. If using the same delivery date, the date saves correctly in the db but the main task list progress is incorrect. If you attempt to set individual delivery dates, the page crashes. 

Need to fix: 

- Quantity status in Solutions and services, 
- Being able to select individual delivery dates, 
- Status of Planned delivery date section in main task list.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/SolutionSelection/QuantityController.cs
ashleywntrbjss: Removed this - now included with query - delivery dates needed it as well

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Controllers/Contracts/DeliveryDatesController.cs
ashleywntrbjss: RolledUp was missing order item (sublocation) recipient copying, so have done that now - that fixed the statuses. Delivery dates was fixed by providing the sublocation name

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Orders/OrderWrapper.cs
ashleywntrbjss: Reverted to previous logic

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderSublocationService.cs
ashleywntrbjss: Removed excessive validation

src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Orders/OrderWrapper.cs
ashleywntrbjss: Reverted to previous logic (with flattened recipients in place of recipients)

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderSublocationService.cs
ashleywntrbjss: Global filter concept working but out of scope for this change so rolled back

src/NHSD.GPIT.BuyingCatalogue.Services/Orders/OrderService.cs
ashleywntrbjss: Implemented action filter attribute on key order actions to prevent access if not editable.

bjss-colin (approved):
A monumental change that'll bring a good performance uplift. Well done

PR #1544, opened 2025-06-23
Use correct hint ID for aria-describedby attribute

bjss-colin:
We use different ID formats depending on the context of the page. In some cases it's an array of elements denoted by an indexer. In other cases it's a single element with a name.

image

This moves the name logic out of the hint label builder as it's the upper tagbuilder that should decide the name to be used.

In cases where multiple hints existed, these were being delimited by an underscore thus resulting in an invalid element ID. For example:
1. Hint label - `my-input-hint`
2. Character count `my-input-summary`

resulted in an aria-describedby value of `my-input-hint_my-input-summary` when it should be space delimited like `my-input-hint my-input-summary`


ashleywntrbjss (approved):
Approved, see ticket for test details

PR #1545, opened 2025-06-27
Set default solution standards when adding a new solution

bjss-colin:
Applies the following changes:

1. Move the Function App into the Web App Solution
2. Map the new "Fully Met - Default" column in the CSV
3. Map the `InProgressStandards` when adding a solution

For compatibility, until the CSV files are updated permanently, the `InProgressStandards` are applied conditionally only when there's at least one Standard in the database that has a `IsMetByDefault` state of true.


ashleywntrbjss (approved):
Approved see ticket for test evidence

PR #1547, opened 2025-07-03
Competition accessibility fixes

bjss-colin:
1. Refactor the CatalogueItemId parsing logic to remove regex (tech debt - regex not needed for simple parsing)
2. Tweak the aria-described by again. This was sanitizing the whole attribute in one pass rather than sanitizing individual IDs and joining them all together into a single attribute value at the end
3. Accessibility improvements (described in story comments)


ashleywntrbjss (approved):
Approved see ticket for test evidence

PR #1549, opened 2025-07-17
Feature/sublocation hub for competitions and orders

ashleywntrbjss:
This PR:

Implements sublocation based recipient selection for the competition and order journeys, which share a common set of components.

Users are now prompted to select their sublocations before their recipients. There is also a hub page where the status of each sublocation can be checked.

The mergers and split order journey continues to use the old components and will be updated in future.

The underlying entities for recipients have been upgraded to a new version that includes the sublocation reference, with new sublocation tables to store the competition / order and sublocation relationship separately.

Recipients are migrated to the new sublocation recipients tables using existing ods relationship data, with some manual intervention for relationships that have since been deactivated.


bjss-colin (approved):

bjss-colin (changes requested):
Looks like `NHSD.GPIT.BuyingCatalogue.WebApp.UnitTests.Areas.Admin.Controllers.ManageOrdersControllerTests.Get_Download_InProgressOrder_ReturnsExpectedResult` has become flaky and test data needs some tweaks to be more concrete

bjss-colin (approved):

PR #1550, opened 2025-07-17
Bugfix: Use price weighted score for competition scoring

bjss-colin:
The competition journey currently uses the full price score + the non-price element weighted score when calculating the winners which results in incorrect results when non-price elements are weighted higher than price (I.E; a user wants non-price to influence the results more than price).

While a score should only be as high as 5, this bug means that a solution score could be set to 7.1 out of 5 and is then picked as the winner.

When the user views the results page, they're then given the weighted scores correctly making it all the more confusing as to why a solution with a lower score won the competition.

The test case reflects the scenario whereby there are joint-winning solutions and one non-winning solution has a higher price score but lower non-price scores.


ashleywntrbjss (approved):
Approved see test evidence 

Before after comparison showing correct solution picked

PR #1552, opened 2025-07-28
Bugfix/24166/waf rule workaround for 'on'

ashleywntrbjss:
This pr:

<strike>Adds a rule exclusion to prevent certain common search terms from triggering the WAF</strike>

The rule couldn't be that precise, so now the request itself gets padded to prevent ?on from being the final parameter in a request


bjss-colin (approved):
It's a shame about the lack of flexibility in WAF rules but this is a sufficient workaround for now

PR #1554, opened 2025-08-01
Calculate cumulative cost for amendments

bjss-colin:
The existing price calculator was originally written for normal orders. It's not compatible with amendments when it comes to tiered pricing because it doesn't track the total quantity and offset the tiers for the subsequent amendments to fit newer quantities into the most applicable tier.

For instance, if an order is placed with a quantity of 760,644 against a product that uses tiered cumulative pricing then the tiers would be assessed as such:

| Tier | Range      | Price   | Quantity |
|------|------------|---------|----------|
| 1    | 0-89,999   | £0.1660 | 89,999   |
| 2    | 90,000-899,999 | £0.1330 | 670645   |
| 3    | 900,000+   | £0.08   | 0        |

When an amendment is placed with an additional quantity of 16,277 then it's being calculated as follows because the calculator does not track the whole order, rather it tracks and calculates individual revisions causing every calculation to start in the first tier.

| Tier | Range      | Price   | Quantity |
|------|------------|---------|----------|
| 1    | 0-89,999   | £0.1660 | 16,277   |
| 2    | 90,000-899,999 | £0.1330 | 0        |
| 3    | 900,000+   | £0.08   | 0        |

This results in a price of `16,277 * 0.1660` which is incorrect. the tier structure should instead be:

| Tier | Range      | Price   | Quantity |
|------|------------|---------|----------|
| 1    | 0-89,999   | £0.1660 | 0        |
| 2    | 90,000-899,999 | £0.1330 | 16,277   |
| 3    | 900,000+   | £0.08   | 0        |

This results in the correct price calculation of `16,277 * 0.1330` since that quantity fits into the remaining capacity of tier 2 from the original order.

For clarity, the total quantity of the whole order (original + amendments combined) would then be

| Tier | Range      | Price   | Quantity |
|------|------------|---------|----------|
| 1    | 0-89,999   | £0.1660 | 89,999   |
| 2    | 90,000-899,999 | £0.1330 | 686,922  |
| 3    | 900,000+   | £0.08   | 0        |

The sum of each order subtotal is combined and multiplied by the term (typically the contract length for the original order and the _remaining_ contract length for the amendment - these run consecutively)

Volume is more complicated since an increase in a tier, as a result of a higher quantity during an amendment, on volume pricing should move all quantities over to the new tier. This means that the term of the order revisions will be the _completion_ date of the next revision while the term of the most recent revision is the contract end date.

Complicating it further is that one additional service could use a volume price while another uses a cumulative price - these need to be calculated based on different term rules. At the moment we calculate all items at once and multiply against the contract end-term.

Due to these complications, I've disabled Volume Tiered pricing for now until we receive all billing logic/rules.


hg-engineer (dismissed):

src/NHSD.GPIT.BuyingCatalogue.Framework/Calculations/CataloguePriceCalculations.cs
```diff
@@ -126,14 +132,55 @@ private static decimal CalculateForTerm(
...
 
-            var price = orderItem.OrderItemPrice as IPrice;
+            IPrice price = orderItem.OrderItemPrice;
             return price.CalculateOneOffCost(orderItem.TotalQuantity(recipients))
-                       + (price.CalculateCostPerMonth(orderItem.TotalQuantity(recipients)) * term);
+                + (price.CalculateCostPerMonth(orderItem.TotalQuantity(recipients)) * term);
         }
 
-        private static decimal TotalCost(this Order order, Order previous)
+        private static decimal TotalMonthlyCost(
```
hg-engineer, line 140: could this method be refactored to include a helper method to reduce the lines of code in the method?

hg-engineer (approved):

src/NHSD.GPIT.BuyingCatalogue.Framework/Calculations/CataloguePriceCalculations.cs
ghost: This will be tweaked again in the future when we include volume based pricing. The plan then is to break it out further into smaller methods

PR #1562, opened 2025-09-04
content improvements - correcting text and capitalisation

hg-engineer:


bjss-colin:
There are many instances where the terminology is converted lowercase when used at the beginning of a sentence. It's worth starting this one again and tackling in smaller batches as 235 files in one sweep is a bit too large to review content/grammar rules.

Let's focus exclusively on the order journey (section 1) and update the HTML files, excluding any C#. In places where enums are used  then we should use `ToLowerInvariant` in the HTML/Model rather than updating the enums

PR #1588, opened 2025-11-20
Migrate mergers and splits to new service recipients sublocations pages

bjss-colin:
Implements the new service recipients journey within the mergers and splits journeys.

The mergers/splits journeys are unique in that one recipient is uniquely marked as the recipient that is being split or retained. In the past, under the flat service recipient structure, this recipient was removed from the `OrderRecipients` table in the database and it was re-added at runtime during the recipients page. This meant that quantities, delivery dates and prices were all handled correctly with respect to the remaining recipients.

The new parent/child structure doesn't work with this approach as both the sublocation and recipient need to be retained. The implemented approach now keeps the special recipient in the table and performs the filtering at runtime to shift the business logic into the application.

The new function `GetOrderRecipients` on the `Order` object performs the filtering on the `FlattenedRecipients` property. Ideally, all references to `FlattenedRecipients` would be replaced with `GetOrderRecipients` but this can be picked up as a subsequent change as it's largely a refactoring.

Previously, we could validate that the required number of service recipients were selected. In this journey the selections happen on separate pages. New logic has been added to redirect to user to the order task list page, with the service recipients step set to "in progress" until at least 2 recipients are selected and the retained/split recipient is selected.


bjss-colin:

PR #1589, opened 2025-11-28
Bugfix: Use product price for validation in order journey

bjss-colin:
There's currently a bug in the ordering journey whereby the order price is used in the validation for the pricing constraints, whereas it should use the original list price to constrain the order price to prevent buyers from exceeding the list price value.

For example, if a user wants to purchase a product with a price of £1.26 and then user has negotiated a better price of £0.01. If the user later tries to change the price from £0.01 to £1.00 then a validation error is thrown as it believes £0.01 is the original price and not £1.26


bjss-colin:

PR #1590, opened 2025-11-28
Rename order summary and remove signature block

bjss-colin:
Remove the signatures block from the order summary as it is an invoice and not a legally binding document


bjss-colin:

PR #1598, opened 2026-01-12
[24233] Bugfix - Incorrect Call-off ID on service recipients

adawari-thomas-bjss:
## Changes
- Updated the previous recipients listed for an amendment to display the call-off id related to its original order

## Validation
image


src/NHSD.GPIT.BuyingCatalogue.ServiceContracts/Orders/OrderWrapper.cs
```diff
@@ -39,6 +39,11 @@ public OrderWrapper(Order currentOrder, IEnumerable<Order> previousOrders)
 
                 Order output = previous.First().Clone();
 
+                foreach (OrderSublocationRecipient recipient in output.FlattenedRecipients)
```
ghost: `FlattenedRecipients` was introduced as a means of convenience around a year ago. More recently I introduced `GetOrderRecipients` which performs some filtering for different order journeys.

For consistency it might be best to use `GetOrderRecipients` here as it'd be good to eventually mark `FlattenedRecipients` as private.

bjss-colin (dismissed):
That's good stuff. Tested and working well across the PDF + website version.

1 optional improvement but I've approved

bjss-colin (approved):

PR #1603, opened 2026-01-22
Add new radio buttons to track supplier explanation requirements

bjss-colin:
# Changes

* Adds a new radio button input to the associated service requirements page
* Validation error message when user tries to submit page without making a selection
* Displays the content in the requirements table
* Displays the content on the requirement deletion page
* Displays the content on the order summary page


adawari-thomas-bjss (approved):
Approved. Minor, but you might want to take a look at the sonar issues

PR #1612, opened 2026-02-17
[Per unit quantities] Reflect service recipient quantity for all price types

bjss-colin:
# Change

Updates the CSV and some intermediary pages to display the newer per service recipient based quantities in a manner that is backwards compatible with the existing orders which use the older order item based quantity


src/NHSD.GPIT.BuyingCatalogue.Services/Csv/CsvService.cs
```diff
@@ -200,19 +200,19 @@ private async Task<List<FullOrderCsvModel>> GetModelListForCsv(int orderId)
                         CallOffId = or.Order.CallOffId,
                         OdsCode = or.Order.OrderingParty.ExternalIdentifier,
                         OrganisationName = or.Order.OrderingParty.Name,
-                        SubIcbCode = !(oir.OrderItem.OrderItemPrice as IPrice).IsPerServiceRecipient()
+                        SubIcbCode = oir.OrderItem.Quantity.HasValue
```
ghost, line 203: Backwards compatibility. If `orderitem->quantity` has been specified on any existing orders then use that

adawari-thomas-bjss (approved):

PR #1619, opened 2026-03-17
Add Call-off terms to order journey

bjss-colin:


bjss-colin:
Force-merged due to developer absence. Change was reviewed on Friday prior to annual leave and the only issue that appeared was relating to the task list progress that was resolved in commit 17c824e

PR #1620, opened 2026-03-17
db schema changes 

adawari-thomas-bjss:
OrderingSchemaProposed


database/NHSD.GPITBuyingCatalogue.Database/PostDeployment/migrateToNewTables.sql
```diff
@@ -0,0 +1,288 @@
...
+    BEGIN TRANSACTION;
+
+    DECLARE @BatchSize INT = 2000;
+    DECLARE @RowsInserted INT = 1;
+
+    WHILE @RowsInserted > 0
+    BEGIN
+      MERGE INTO ordering.OrderItemsV2 AS TARGET
+      USING (
+        SELECT TOP (@BatchSize)
```
ghost: The batch size is never incremented so this will only write the top 2,000 rows

We can remove the batch size and `@RowsInserted` logic as they shouldn't be necessary with how little data we're working with.

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Extensions/CollectionExtensions.cs
```diff
@@ -39,14 +39,12 @@ public static bool AllQuantitiesEntered(this ICollection<OrderSublocationRecipie
 
     public static bool SomeButNotAllNewQuantitiesEntered(
         this ICollection<OrderSublocationRecipient> recipients,
-        OrderItem orderItem,
```
ghost, line 42: Looks like we're removing amend specific logic. Is this residing elsewhere now?

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Ordering/Configuration/OrderItemEntityTypeConfiguration.cs
```diff
@@ -30,17 +29,24 @@ public void Configure(EntityTypeBuilder<OrderItem> builder)
...
-                .HasConstraintName("FK_OrderItems_Order");
+                .HasConstraintName("FK_OrderItemsV2_Order");
 
             builder.HasOne(oi => oi.LastUpdatedByUser)
                 .WithMany()
                 .HasForeignKey(oi => oi.LastUpdatedBy)
-                .HasConstraintName("FK_OrderItems_LastUpdatedBy");
+                .HasConstraintName("FK_OrderItemsV2_LastUpdatedBy");
+
+            builder.HasOne(oi => oi.Parent)
```
ghost, line 44: Nicely done

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Ordering/Models/OrderItem.cs
```diff
@@ -41,11 +48,14 @@ public OrderItem(
...
         public OrderItemPrice OrderItemPrice { get; set; }
 
+        public ICollection<OrderItem> Services { get; set; } = new HashSet<OrderItem>();
+
         public OrderItem Clone() => new(CatalogueItemId)
         {
             OrderItemPrice = OrderItemPrice?.Clone(),
             EstimationPeriod = EstimationPeriod,
             CatalogueItem = CatalogueItem,
+            CatalogueItemId = CatalogueItemId,
```
ghost, line 56: Let's double check that all required elements are cloned for any amendments that are created. Worth doing some A-B testing between the demo environment and locally to compare the order summary and PDF pages

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Ordering/Models/OrderItemSublocationRecipient.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel.DataAnnotations.Schema;
```
ghost: Looks like an unused import. This could be removed

src/NHSD.GPIT.BuyingCatalogue.Services/Competitions/CompetitionOrderService.cs
```diff
@@ -120,27 +123,43 @@ public async Task<CallOffId> CreateOrder(string internalOrgId, int competitionId
...
+        var orderItems = new List<OrderItem>() { solutionOrderItem };
+        var associatedServicesOrderItems = directAwardSolution.GetAssociatedServices()
+            .Select(x => new OrderItem(x.CatalogueItemId) { Created = DateTime.UtcNow })
+            .ToList();
+        var additionalServicesOrderItems = directAwardSolution.GetAdditionalServices()
+            .Select(x => new OrderItem(x.CatalogueItemId) { Created = DateTime.UtcNow })
             .ToList();
 
-        orderItems.Add(new OrderItem(directAwardSolution.CatalogueItemId) { Created = DateTime.UtcNow });
+        solutionOrderItem.Services.AddRange(associatedServicesOrderItems);
```
ghost: We add the associated service here and to the list of order items, which the solution is part of. Under the new schema, given the parent will be the solution, should the services both be added to the solution items rather than the top-level list?

src/NHSD.GPIT.BuyingCatalogue.Services/Competitions/CompetitionOrderService.cs
```diff
@@ -120,27 +123,43 @@ public async Task<CallOffId> CreateOrder(string internalOrgId, int competitionId
...
+            winningSolution.Price.Tiers);
+        winningSolutionOrderItem.Services.AddRange(associatedServicesOrderItems);
 
-        orderItems.Add(
-            CreateOrderItem(
-                winningSolution.CatalogueItemId,
-                winningSolution.Quantity,
-                winningSolution.Price,
-                winningSolution.Price.Tiers));
+        var orderItems = new List<OrderItem>() { winningSolutionOrderItem };
```
ghost, line 148: Same as above

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Models/Contracts/DeliveryDates/OrderItemRecipientModel.cs
```diff
@@ -9,10 +9,10 @@ public OrderItemRecipientModel()
         {
         }
 
-        public OrderItemRecipientModel(OrderSublocationRecipient recipient, CatalogueItemId catalogueItemId)
+        public OrderItemRecipientModel(OrderSublocationRecipient recipient, OrderItem orderItem)
```
ghost: We should still be able to pass `CatalogueItemId` here and in similar instances where it's the only thing that's used.

src/NHSD.GPIT.BuyingCatalogue.WebApp/Areas/Orders/Views/Order/Summary.cshtml
```diff
@@ -11,8 +11,8 @@
 <partial name="Partials/_BackLink" model="Model" />
 
 @{
-    var solutions = Model.RolledUp.GetSolutions().ToList();
-    var additionalServices = Model.RolledUp.GetAdditionalServices().ToList();
+    var solutions = Model.Order.GetSolutions().ToList();
```
ghost: As we've changed from `RolledUp` to `Order`, let's do some testing between demo and local to ensure the order journey behaves the same across original orders and amendments

bjss-colin (changes requested):
A few comments and observations as well as some SonarQube warnings

src/NHSD.GPIT.BuyingCatalogue.EntityFramework/Extensions/CollectionExtensions.cs
adawari-thomas-bjss: The previous logic does not work anymore as the `DetermineOrderRecipients` method is used to get the recipients list that calls this extension method. The recipient list is no longer an aggregation of the recipients across each revision, therefore, we don't need to keep track of the previous recipients

src/NHSD.GPIT.BuyingCatalogue.Services/Competitions/CompetitionOrderService.cs
adawari-thomas-bjss: The solution item and the additional service would both be top-level items as they both can have additional services, so I don't think we should add both to the services of the solution item. I'm treating the top-level list similar to a basket to keep pricing calculations as they are, while using the parentId/services for referencing and navigation. But we can discuss a different approach to this

ImranBashir-cgi (approved):

ghost (approved):

ghost (approved):

PR #1622, opened 2026-03-24
Allow solution creation on inactive suppliers

bjss-colin:


bjss-colin:
Force-merge due to developer absence. Signed off my product manager

PR #1623, opened 2026-03-24
Terraform for call-off terms url

bjss-colin:


bjss-colin:
Force merged due to developer absence. Adds terraform required for #1619

PR #1624, opened 2026-03-25
Calculate one-off cost of all competition services

bjss-colin:
Resolves an issue where Additional Services were not factored in the one-off cost calculations


tests/NHSD.GPIT.BuyingCatalogue.EntityFramework.UnitTests/Models/Competitions/CompetitionSolutionTests.cs
adawari-thomas-bjss, line 1: Could there be a test case that covers a scenario with an additional service and an associated service?

ghost: Good shout. New test added to cover a scenario where the associated service is a one-time payment, the additional service is billed monthly and solution yearly

adawari-thomas-bjss (approved):

PR #1637, opened 2026-04-30
Feature/playwright test framework

ImranBashir-cgi:
This PR introduces the core infrastructure for the Playwright end-to-end test framework. It establishes the foundation that all future page objects and tests will build upon.


tests/NHSD.GPIT.BuyingCatalogue.PlaywrightTests/Infrastructure/TestServerFixture .cs
```diff
@@ -0,0 +1,125 @@
+﻿using System.Diagnostics;
+using System.Text;
+
+namespace NHSD.GPIT.BuyingCatalogue.PlaywrightTests.Infrastructure
+{
+    public class TestServerFixture : IAsyncLifetime
+    {
+        private static readonly string ProjectPath = Path.GetFullPath(
```
ghost: For the regression suite we should assume that the application will already be running in the background. There should be a URL parameter that is passed to the test suite during runtime which will be the URL of the environment.

Unfortunately, this approach won't work because the csproj won't exist in the release pipeline. In fact, the tests will be running on an entirely different server to the application itself.

For any local tests, the team will have to run the web app first and then run the tests afterwards.

ImranBashir-cgi: Updated TestServerFixture to remove process management entirely. BaseUrl is now read from the TEST_BASE_URL environment variable at runtime, falling back to https://localhost:5001 for local development.

ghost (approved):

